PR #22, opened 2017-01-02
add response view options

vincent178:


Romakita:
Hi vincent,
Thanks for your commit :) . Your feature is available with the v1.3.0-8.

Happy new year !

PR #24, opened 2017-01-11
[Decorator] Add table and model related decorators

vincent178:
Hi @Romakita ,

I just added two new decorators, one is `@Table`, one is `@Column`.

These two mainly for build db query, the Table will accept one parameter which is the table name in database.

The second one have an optional parameter to which is the column name in the database.

I also add one abstract model, all the application level model should extend this abstract model.
And every application level model have two static methods: `table`, `columns` then.

For example:

```
class Comment extends Model {
    @Column("_id")
    private id: number;
}

Comment.table() # => "comments"
Comment.columns() # => ["comments._id"]
```

So I can build the query with Knex like follows
```
.select(Comment.columns())
.from(Comment.table())
.where(...)
```

### DIFF
--- a/src/constants/metadata-keys.ts
+++ b/src/constants/metadata-keys.ts
@@ -1,33 +1,37 @@
 // controllers
-export const CONTROLLER_URL = "tsed:controller:url";
-export const CONTROLLER_DEPEDENCIES = "ted:controller:depedencies";
-export const CONTROLLER_MOUNT_ENDPOINTS = "ted:controller:endpoints";
-export const ENDPOINT_ARGS = "tsed:endpoint";
-export const ENDPOINT_VIEW = "tsed:endpoint:view";
-export const ENDPOINT_VIEW_OPTIONS = "tsed:endpoint:view-options";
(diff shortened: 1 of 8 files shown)


vincent178:
Hi @Romakita , could you please help check the error travis one?

vincent178:
ok, after added some tests, the travis passed, and the test coverage increased. :)

Romakita:
Hi @vincent178 ,

Thanks for your commit. It's great idea. But I think theses decorators are out of the  "ts-express-decorators" scope. If I say that it's only to keep a certain coherence and not to question your idea ;)

I have the same needs with MongoDB but I didn't implement it in ts-express-decorators. 

Maybe we need to create another module ? 

Anyway, it's a good job :)

See you,
Romain

vincent178:
Hi @Romakita 

I think you don't need to merge it util we have an agreed plan.
For now, how about we just keep it here?

Thanks,
Vincent

Romakita:
Hi @vincent178 ,

Yes you can keep. I'll merge your commit after the next release ;).

See you,
Romain

PR #71, opened 2017-06-19
Feature/#26 open api spec genration

vologab:


Romakita:
Hi @volobag,

Thanks a lot for your PR :). I'll merge soon your code on the master branch and release a new version with your feature!

See you

vologab:
Hi @Romakita,
To be honest this feature it's not complete enough, but it works:
You can add to RestCtrl inside [Basic Example](https://github.com/Romakita/example-ts-express-decorator/tree/1.4.0/basic) method smth like that:
``` javascript
    @Get('/swagger.json')
    public getOpenAPISpec() {
        return this.routeService.getOpenAPISpec();
    }
```

And then add lib https://github.com/scottie1984/swagger-ui-express for adding Swagger-UI middleware.

```javascript
// ... inside Server.ts $onMountingMiddlewares
swaggerUiMiddleware = require('swagger-ui-express');
....
this..use('/api-docs', swaggerUiMiddleware.serve, swaggerUiMiddleware.setup({}));
```

Then you will be able use Swagger UI for exploring your API with URL:  http://localhost:8080/api-docs/?url=/api/swagger.json

If you want, I can add this to Basic Example, after feature will be merged to master?

vologab:
@Romakita Could you please clarify how we can achieve some important things from Decorators for OpenAPI spec?
We need following data:
 * Info object https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-schema-official/index.d.ts#L6
* Consumes property https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-schema-official/index.d.ts#L221
 * Security object https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-schema-official/index.d.ts#L205
 * Response object https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/swagger-schema-official/index.d.ts#L109

Romakita:
Hi @vologab;

yes you can create pull request on the basic example or create a new example name example-swagger-ui ^^.

For your second question. I think we need to create decorators for theses points:
- Info object,
- Consume property,
- Security object,

But I'm not sure about the data requirement for each object. May be with some example I can give you the good method to find the data in ts-express-decorators.

For the Response Object. I'm not sure. Typescript generate metadata on the method. But when a method return a Promise<Type>, the Type is lost. So ... maybe we need a decorator too.

I'll create a branch with your proposal and move some code into a new SwaggerService to separate feature from the core code (because v1.5 change some classes and internal API).

Anyway Good job :)
See you,
Romain

PR #109, opened 2017-09-14
Add Store to Index

lwallent:
In the documentation the import is described as:
import {Store, UseBefore} from "ts-express-decorators";
Hence the Store should be in the Index?


Romakita:
Thanks @lwallent :)

PR #110, opened 2017-09-14
importHelpers true but no lib dependency?

lwallent:
It seems importHelpers was added to tsconfig.json but no dependency on tslib was added to package.json?
This seems like a natural dependency else the build will fail.


Romakita:
Thanks @lwallent for your help :)

PR #112, opened 2017-09-18
Endpointmetadata always includes controller level

lwallent:
fix(endpointmetadata): Endpointmetada on controller is always considered

Previously the controller metadata was only used if the key was not found on the endpoint itself.
This made it impossible to place shared metadata on the controller level.
Example: a common JWT token annotation can now add a 401 response on the controller level which is merged with the other responses added at the endpoint level.


Romakita:
Thanks @lwallent :)

I haven't seen that ;). Great job.

PR #123, opened 2017-09-24
move swagger-schema-official to dev dependencies, use '@types/swagger…

vologab:


vologab:
Sorry, seems we need revert this changes, for this reasons:
1) https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html
2) Similar issue :) https://github.com/Polymer/dom5/issues/45

Romakita:
No problem. But when you move a @types files to devDependencies (in this case, in ts-express-decorators module) you must install the dependencies (@types/swagger) explicitly in your project.

This is why, in the getting started, the prerequisite is to install `@types/express`.

I'll revert the change and set @types/swagger to the latest version ;)

Issue #128, opened 2017-09-26
Swagger model contains all information from all classes that inherit a base model


Ahagan:
## Information

Type | Version
---|---
Bug | 2.3.8

****

## Description
If a model class inherits a base class that also contains the swagger decorators (e.g. @Title('')) then when the swagger documentation is generated, the model contains details about all properties from all models that inherit from the base class.

E.g.

Base Model ->
```typescript
export abstract class BaseModel {
    @Title('id')
    @Description('Unique identifier.')
    @JsonProperty()
    public id?: string;
}
```
Models ->
```typescript
export class NameModel extends BaseModel{
    @Title('name')
    @Description('The name')
    @JsonProperty()
    public name: string;
}

export class AgeModel extends BaseModel{
    @Title('age')
    @Description('The age')
    @JsonProperty()
    public age: number;
}
```

output model 
- `id`: Unique identifier
- `name`: The name
- `age`: The age

## Todos

- [ ] Fix swagger documentation generation if a class inherits from a base class with the swagger decorators on them.

Romakita:
@Ahagan Thanks for this issues :) I'll fixed that ASAP ;)

Romakita:
@Ahagan  Fixed :) with 2.4.1

PR #131, opened 2017-09-30
fix(core,mvc,converter): Fix inheritance metadata

Romakita:
## Informations

Type | Status | Migration
---|---|---
Fix | Ready | No

****

## Description
Converter can serialize and deserialize class which have a parent classes.
Swagger can use metadata stored on model which have a parent classes.

Closes: #128 

## Todos

- [x] Tests
- [x] Coverage


src/converters/registries/PropertyRegistry.ts
```diff
@@ -16,24 +17,43 @@ export class PropertyRegistry {
...
         }
 
-        return this.getProperties(target).get(propertyKey) as PropertyMetadata;
+        return this.getOwnProperties(target).get(propertyKey) as PropertyMetadata;
     }
 
     /**
      *
      * @param target
      * @returns {Array}
```
lwallent: The return type in the documentation is not correct ;-) These are always tricky to keep updated.

src/converters/registries/PropertyRegistry.ts
```diff
@@ -16,24 +17,43 @@ export class PropertyRegistry {
...
      */
-    static getProperties = (target: Type<any>): Map<string | symbol, PropertyMetadata> =>
-        Metadata.has(PROPERTIES_METADATA, target)
-            ? Metadata.get(PROPERTIES_METADATA, target)
+    static getProperties(target: Type<any>): Map<string | symbol, PropertyMetadata> {
+        const map = new Map<string | symbol, PropertyMetadata>();
+        let currentTarget = getClass(target);
+
+        while (nameOf(currentTarget) !== "") {
```
lwallent: Is this really the desired traversal order? Do we not want depth first traversal or am I missing the point? 
Will this not let the parent overwrite the children? In case the property is present in multiple places in the hierarchy will this produce the correct result? I would assume starting at the root and then set the keys coming back down?

Romakita: ho yes!! I missed that... Thanks @lwallent , I'll rewrite this ;)

src/converters/registries/PropertyRegistry.ts
vologab: May be https://github.com/TypeStrong/typedoc could help in this cases?

Romakita: The problem with Typedoc the generator doesn't support the latest typescript version. And I don't like the template, not clear, not easy to use.

I'll fix the documentation ;)

lwallent:
Looks fine to me 👍

Issue #135, opened 2017-10-05
Ability to overwrite initial request log


Ionaru:
## Information

Type | Version
---|---
Improvement | 2.4.1

****

## Story
As a developer, I want to be able to conditionally disable the logging of requests so that I can keep requests to images, JS files, CSS files, etc. out of the log.

## Possible solution
Change the current implementation in `LogIncomingRequestMiddleware.ts`
```typescript
public use(@Req() request: Express.Request, @Res() response: Express.Response): void {

    this.configureRequest(request);

    request.log.info();

    applyBefore(response, "end", () => this.onLogEnd(request, response));
}
```

To something like this
```typescript
public use(@Req() request: Express.Request, @Res() response: Express.Response): void {

    this.configureRequest(request);

    this.onLogStart(request);

    applyBefore(response, "end", () => this.onLogEnd(request, response));
}

protected onLogStart(request: Express.Request): void {
     request.log.info();
}
```

The seperate `onLogStart()` function will allow me to overwrite the initial request log and to add my own logic to it.

Romakita:
Hi @Ionaru ,

You can create PR ;)
See you,
Romain

Ionaru:
I'm not really up-to-speed on the dev environment, documentation and testing of Ts.ED, thought it would be faster to make an issue instead. :wink:

Romakita:
ok :)

Romakita:
@Ionaru  PR ready : https://github.com/Romakita/ts-express-decorators/pull/136

PR #136, opened 2017-10-05
fix: Refactoring log start.

Romakita:
## Informations

Type | Status | Migration
---|---|---
Refactoring | Ready | Yes

****

## Description
Move log start into a method to facilitate the override logger.
Closes: #135

## Todos

- [x] Tests
- [x] Coverage

### DIFF
--- a/src/mvc/components/LogIncomingRequestMiddleware.ts
+++ b/src/mvc/components/LogIncomingRequestMiddleware.ts
@@ -5,12 +5,12 @@
 import * as Express from "express";
 import {$log} from "ts-log-debug";
 import {EnvTypes} from "../../core/interfaces";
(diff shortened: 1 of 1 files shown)


src/mvc/components/LogIncomingRequestMiddleware.ts
```diff
@@ -36,11 +36,20 @@ export class LogIncomingRequestMiddleware implements IMiddleware {
         this.configureRequest(request);
 
         request.log.info();
```
Ionaru: Oops! I didn't notice this, sorry. :cry: 

The original `request.log.info();` should be removed.

Romakita: Fixed ;)

Issue #130, opened 2017-09-28
@Required() does not validate @BodyParams


Ahagan:
## Informations

Type | Version
---|---
Bug | 2.3.8

****

## Description
When @Required() is added before @BodyParam(), it is expected that an exception would be thrown if no body params were posted to the endpoint or if the body params did not match the expected type.

E.g.

    @Post('/')
    async broadcast(@Required() @BodyParams() webhook: Webhook): Promise<any> {
        ,,,
    }

and the models is
export class WebHook {
   url: string;
  name: string;
}

- if I posted up {}, I'd expected the Required decorator to throw an exception.
- if i posted up { a: '', b: ''} I'd expect the Required decorator to throw an exception because the properties do not match what is on the WebHook class.
- if i posted up no body, I'd expect the Required decorator to throw an exception
- if i posted up { url: 'http://google.co.uk', name: 'test'}, I'd expect the Required decorator to not throw an exception
- if the body params has nested objects these should also be checked that they have been set.

At the moment Required does not validate the BodyParams at all.

## Todos

- [ ] Update Required decorator to validate the BodyParams

Romakita:
Hi @Ahagan,
It's a part of these stories #85 and #35 .

Actually Required work fine only with simple data like this:

```typescript
@Required @BodyParam('name') name: string;
```
With a model, we need to implement a model validators. It's planed :)
See you

Ahagan:
Ok, thanks for letting me know. Do you know when the work will be started on this?

Romakita:
October, I hope ^^.

virus2016:
Hey @Romakita,

Is it possible to add the class validation library now for existing apps? Then once it’s implemented then just move back?

I would love to help out on this but just don’t seem to have time. :-(

Just a thought!

Ahagan:
@Romalita thanks, and keep up all the great work! 😊

Romakita:
@Ahagan the story for the validation feature is ready. You can follow this issue #85

Issue #35, opened 2017-01-15
[Decorator] Improve Required() decorator on Model deserialization


Romakita:
## Status
**Enhancement**

## Description
As a developer, I want to be able to Required decorator on a model Property. 

## Usage example
```typescript
class Model {
   @Required()
   name: string;

   @Required()
   @Allow(null)
   obj: Object;
}

class ResponseModel {
   @Required()
   _id: string;

   @Required()
   name: string;

   @Required()
   @Allow(null)
   obj: Object;
}

class MyController {
    @Post("/")
    async save(@Required() @BodyParams() model: Model): Promise<ResponseModel> {
       return new ResponseModel() // will use to validate the structure and required value
    }
}
```
## Acceptance criteria

- [x] Required can be used on model property
- [x] Required add bad request on swagger documentation
- [x] Model can be used on return type method
- [x] Allow decorator create value exception when Required is used

PR #138, opened 2017-10-07
feat(mvc): Improve the Required decorator

Romakita:
## Informations

Type | Status | Migration
---|---|---
Feature | Ready | No

****

## Description
- Required can be use on property
- add Allow decorator

Closes: #130, #35

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation

### DIFF
--- a/src/swagger/class/OpenApiParamsBuilder.ts
+++ b/src/swagger/class/OpenApiParamsBuilder.ts
@@ -2,7 +2,7 @@
  * @module swagger
  */
 /** */
-import {BaseParameter, BodyParameter, Parameter, Schema} from "swagger-schema-official";
+import {BaseParameter, BodyParameter, Parameter, Response, Schema} from "swagger-schema-official";
 import {Type} from "../../core/interfaces";
 import {deepExtends} from "../../core/utils";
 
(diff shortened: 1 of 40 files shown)


src/swagger/class/OpenApiParamsBuilder.ts
```diff
@@ -80,14 +81,17 @@ export class OpenApiParamsBuilder extends OpenApiPropertiesBuilder {
...
     private createBaseParameter(inType: string, param: ParamMetadata): BaseParameter {
         let baseParam: BaseParameter = {
             name: (inType === "body") ? "body" : <string>param.expression,
             in: inType,
             required: !!param.required,
             description: ""
         };
+
+        if (param.required) {
+            this._responses[400] = {description: "Bad request"};
```
lwallent: I would consider changing the description as 400 already says it is a bad request :-)
Why not say "Missing required parameter"?

src/swagger/class/OpenApiPropertiesBuilder.ts
```diff
@@ -61,6 +62,10 @@ export class OpenApiPropertiesBuilder {
             schema.description = schema.description || model.store.get("description");
         }
 
+        if (model.required) {
+            this._responses[400] = {description: "Bad request"};
```
lwallent: Ditto :-)

src/swagger/class/OpenApiParamsBuilder.ts
Romakita: I'm agree with the proposed description ;)

But 422 ? You have an idea ?

Romakita:
Thanks @lwallent for the code review ;)

PR #139, opened 2017-10-08
feat(mvc,swagger): Improve headers and status decorators.

Romakita:
## Informations

Type | Status | Migration
---|---|---
Feature | Ready | No

****
## Description
- Headers is documented in swagger along with default status code,
- add Return decorator,
- refactoring some keys.

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation


src/mvc/decorators/header.ts
```diff
@@ -1,45 +1,82 @@
...
- * @module common/mvc
- */
-/** */
+import {IResponseHeader} from "../interfaces/IResponseHeader";
+import {IHeadersOptions, IResponseHeaders} from "../interfaces/IResponseHeaders";
+import {mapHeaders} from "../utils/mapHeaders";
 import {UseAfter} from "./method/useAfter";
 
 /**
  * Sets the response’s HTTP header field to value. To set multiple fields at once, pass an object as the parameter.
```
lwallent: 👍  +1 for the added documentation :-)

src/mvc/decorators/header.ts
```diff
@@ -1,45 +1,82 @@
...
+ * })
+ * private myMethod() {}
+ * ```
+ *
+ * This example will produce the swagger responses object:
+ *
+ * ```json
+ * {
+ *   "responses": {
+ *     "204": {
```
lwallent: 👍 +1 for the added documentation :-)

Romakita:
Yes, Return is to overkill. I just hesitate to leave in the PR ^^.

I've added this decorator because @Header and @Status now add a documentation on the default response.

Example:
```typescript
@Status(202, {description: "OK"})
@Header('Content-Type', 'text')
myMethod(): Promise<Model>
```
This example will produce the following swagger responses:
```json
{
   "202": {
      "description": "OK",
      "headers": {
          "Content-Type":  {
              "type": "string",
          }
      }
   } 
}
```
> Header information will be automatically related to the 204 status.

But in this case the response type is missing. So actually, we have two choices, to add this information:
```typescript
@Status(202, {description: "OK", type: "Model"}) // here
@Header('Content-Type', 'text')
@Returns(202, {type: "Model"})  // or here
myMethod(): Promise<Model>
```
Status is not bad solution. But use Returns to set the right type with the status code is redundant, no ?

This is why have added Return decorators. 

```typescript
@Status(202, {description: "OK", type: "Model"}) // here
@Header('Content-Type', 'text')
@Return(Model)
myMethod(): Promise<Model>
```
But this solution don't satisfy me. As you said, is more confusing.

Maybe, the good solution is to use Returns decorators without statusCode in first parameters.

```typescript
@Status(202, {description: "OK", type: "Model"}) // here
@Header('Content-Type', 'text')
@Returns(Model) //or @Returns({type: Model})
myMethod(): Promise<Model>
```
What is the best choice ?

See you,
Romain

lwallent:
I personally think the `@Returns` decorator reads better when looking at the code. I do not see myself adding `@Status` decorators all over the place if we are talking about the standard code 200!?

Number one would of course be to read the return type from the actual code but as that is not part of the emitted metadata one should settle for the next best thing. To me the next best thing is using the `@Returns` decorator - as it is describing what should be returned (if successful).

So in short I vote for using `@Returns` :-)

However, I would love to be able to write:

```typescript
 @Returns(Asset[])
```
(where 200 would just be the default value - should not be written in a status)

instead of 

```typescript
 @Returns(200, {use: Asset, collection: Array})
```
I am just dreaming away. Why do you think we need the `@Status` to begin with?

Kind Regards
Lasse

Romakita:
By default, the response will build with the status code 200. Status isn't needed in most case.

```typescript
@Header('Content-Type', 'text')
myMethod(): Promise<Model>
```
produce:
```
{
   "200": {
      "description": "OK",
      "headers": {
          "Content-Type":  {
              "type": "string"
          }
      }
   } 
}
```
I agree,  I would love to be able to write:
```typescript
 @Returns(Asset[])
```
But Typescript didn't permit that because this notation it's a Type and not a Value.
Today, the reason why you need to write explicitly the use and collection it's because, Typescript didn't store all types in metadata.
Example:
```typescript
@Returns(200, {use: EventModel, collection: Array})
async update(@BodyParams("event", EventModel) event: EventModel[]): Promise<EventModel[]>  
    return event;
}
// Javascript output
tslib_1.__decorate([
        returns_1.Returns(200, { use: Event_1.EventModel, collection: Array }),
        tslib_1.__param(0, index_1.BodyParams("event", Event_1.EventModel)),
        tslib_1.__metadata("design:type", Function),
        tslib_1.__metadata("design:paramtypes", [Array]),
        tslib_1.__metadata("design:returntype", Promise)
], EventCtrl.prototype, "update", null);
```
The stored metadata type for items is Array and  for the return type is Promise.

Romakita:
So I'll remove the Return decorators and improve the Returns decorators to accept a shorter syntax:

```typescript
@Returns(Model)  // use status 200 or @Status by default
@Returns(Model, {description: "", collection: Array})
@Returns({description: "", use: Model})
```
And I propose a second decorator:
```typescript
 ReturnsArray(Model) 
 ReturnsArray(Model, {description}) 
```
What do you think about the ReturnsArray ?

lwallent:
👍  sounds good!

`@ReturnsArray(Model)`  would be really sweet!
KR
Lasse

Issue #142, opened 2017-10-11
Cannot set port in ServerSettings to environment variable on azure app service


Soulskater:
Trying to use this config:
`port: process.env.PORT || 8080`
This works on local or heroku for instance, but in azure the enviroment port variable is something like \\.\pipe\e8e054ce-f7b0-4903-9af4-6ceba48a2285

So i get runtime error on app start because there is a validation or something, it says the port should be a number.

Romakita:
Hi @Soulskater 

I'm not sure but can you try to override the startServer and redeploy on azure. Here the fix:

```typescript

class Server extends ServerLoader {
    protected startServer(http: any, settings: { https: boolean, address: string, port: string | number | false }) {
        const {address, port, https} = settings;

        $log.debug(`Start server on ${https ? "https" : "http"}://${settings.address}:${settings.port}`);
        const promise = new Promise((resolve, reject) => {
            http
                .on("listening", resolve)
                .on("error", reject);
        })
            .then(() => {
                $log.info(`HTTP Server listen on ${https ? "https" : "http"}://${settings.address}:${settings.port}`);
            });

        http.listen(port, address);
        return promise;
    }
}
```

Said me if it work and I'll publish a fix ASAP ;)

See you,
Romain

Soulskater:
Seems this fix works!
Could you update the package?
Thanks!

lwallent:
Hi @Soulskater

You should be able to specify the port the same way on Azure.
````
port: process.env.PORT || 8080
````
should work fine.

Have you setup your web.config correctly for iisnode to run the node process?

What is printed in your log during startup?

KR
Lasse

Soulskater:
Here is the error
`ERROR] [TSED] - RangeError: "port" argument must be >= 0 and < 65536
    at assertPort (internal/net.js:17:11)
    at Server.listen (net.js:1393:5)
    at ServerLoader.startServer (D:\home\site\wwwroot\node_modules\ts-express-decorators\src\server\components\ServerLoader.ts:300:14)
    at Server.<anonymous> (D:\home\site\wwwroot\node_modules\ts-express-decorators\src\server\components\ServerLoader.ts:314:32)
    at step (D:\home\site\wwwroot\node_modules\tslib\tslib.js:129:27)
    at Object.next (D:\home\site\wwwroot\node_modules\tslib\tslib.js:110:57)
    at D:\home\site\wwwroot\node_modules\tslib\tslib.js:103:75
[...]

PR #143, opened 2017-10-12
fix(server): fix port when it used on azure server

Romakita:
## Informations

Type | Status | Migration
---|---|---
Fix | Ready | No

****

## Description

Azure set a bitstream in process.env.port. When it used with Ts.ED,
the port is casted to a number. As result, Express receive a NaN object
insteaof a bitstream.

Closes: #142

## Todos

- [x] Tests
- [x] Coverage


Romakita:
@Soulskater,

Here the PR for the fix. Could you review this PR please ? :)

Thanks,
Romain

src/server/components/ServerLoader.ts
```diff
@@ -284,7 +284,7 @@ export abstract class ServerLoader implements IServerLifecycle {
      * @param settings
      * @returns {Promise<TResult2|TResult1>}
      */
-    protected startServer(http: any, settings: { https: boolean, address: string, port: string | number | false }) {
+    protected startServer(http: any, settings: { https: boolean, address: string, port: number }) {
```
Soulskater: Do you really need the port as a number in ts? Just asking because obviously its not cause any runtime error, but it hides the case when i pass the bitstream for instance

Romakita: Yes because Http.listen() waiting a number. And bitstream given by azure it's a hack. In your case, it's not a problem, because the type is lost at the runtime ;)

lwallent: Leaving string in there would be preferable as the named pipe used on Azure is a string.

listen(..) does take a string (path rather that port) where path is a string.
Azure uses the named pipe to map to the right Web Worker. You can look up your "PORT" through Kudu or the like.

👍  @Romakita  +1 for once again being super fast in solving issues :-) 

KR
Lasse

Soulskater: @Romakita +1 from me as well for the awesome fast issue solving! :-)

Romakita: @lwallent ok, I'll leave the string ;)

Romakita:
v2.6.1 @Soulskater

Issue #146, opened 2017-10-19
Model properties no longer compiled to javascript


Ahagan:
## Information

Type | Version
---|---
Bug | 2.6.4

****

## Description
Since upgrading to 2.6.4 if a model is required (as part of BodyParams) does not contain any decorators, the model's properties do not get compiled into JavaScript resulting in an error message when posting a body.

E.g.

Example model:
```typescript
export class WebHookConfig {
    public id?: string;
    public eventType: string;
    public url: string;
}
```

Example controller endpoint:
```typescript
@Controller('/mycontroller')
@MergeParams()
export class MyController {

    public constructor() {
    }

    @Post('/')
    @Responses('404', { description: 'Not found' })
    @Responses('200', { description: 'Success' })
    @Responses('500', { description: 'Error: Internal Server Error' })
    @Summary('EXAMPLE OF ISSUE')
    async post(
        @Required() @BodyParams() model: WebHookConfig
        ): Promise<any> {
        // Method implementation
    }
}
```
When posting up the body I would receive a BadRequest saying that the properties are not allowed on WebHookConfig. As soon as the decorators are added to the model (Title, Description, JsonProperty). I am able to post the data. 

When I look at the output JavaScript, the properties are not added if there are no decorators:
```javascript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
class WebHookConfig {
}
exports.WebHookConfig = WebHookConfig ;
```

When I add the decorators:

```javascript
"use strict";
var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
    var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
    if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
    else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
    return c > 3 && r && Object.defineProperty(target, key, r), r;
};
var __metadata = (this && this.__metadata) || function (k, v) {
    if (typeof Reflect === "object" && typeof Reflect.metadata === "function") return Reflect.metadata(k, v);
};
Object.defineProperty(exports, "__esModule", { value: true });
const lib_1 = require("ts-express-decorators/lib");
const swagger_1 = require("ts-express-decorators/swagger");
[...]

PR #147, opened 2017-10-21
feat(converter): Add global option validationModelStrict.

Romakita:
## Informations

Type | Status | Migration
---|---|---
Featuren | Ready | No

****

## Description
Since v1.6, [ConverterService](http://romakita.github.io/ts-express-decorators/docs/converters.md)
check the consistency between the model and the Json object given to the endpoint.
When a property is unknown on the model, Ts.Ed throw a BadRequest.
You can revert it with validationModelStrict options (ServerSettings).

Closes: #146


Romakita:
@Ahagan

PR #151, opened 2017-10-24
chore(swagger): add a unit test

Romakita:
## Informations

Type | Status | Migration
---|---|---
chore | ready | Yes

****

## Description

Just a complementary test on the generated swagger.json.


test/integration/data/swagger.spec.json
```diff
@@ -0,0 +1,846 @@
...
+        "consumes": [],
+        "responses": {
+          "200": {
+            "description": "Success"
+          },
+          "403": {
+            "description": "Forbidden"
+          }
+        },
+        "security": {},
```
lwallent: Hi @Romakita 
👍 +1 for introducing a unit test!
`security` is an array. I know I added "{}" as the default value yesterday, but I just reintroduced the lines that had been previously removed without really thinking about it :( , sorry about that. Can I make you update the default value as part of this pull request? That is:

```typescript
     const security = this.endpoint.get("security") || [];
```
I think this test was needed :-)
I have not encountered the default value as I have the security set everywhere :-)

Thanks for all your work :-)

Take care
/Lasse

PR #162, opened 2017-11-06
chore(logging): moved to request logging to debug instead of info

lwallent:
chore(logging): moved the request logging to debug instead of info

In an attempt to reduce the ordinary logging the request logging was moved to debug level

@Romakita 
Maybe the real solution would be to split out the request logging in an info part and a debug part. However including the sometimes extreme amounts of data in the info part renders my info logging level worthless :-(

### DIFF
--- a/src/mvc/components/GlobalErrorHandlerMiddleware.ts
+++ b/src/mvc/components/GlobalErrorHandlerMiddleware.ts
@@ -26,7 +26,7 @@ export class GlobalErrorHandlerMiddleware implements IMiddlewareError {
         const toHTML = (message = "") => message.replace(/\n/gi, "<br />");
 
         if (error instanceof Exception || error.status) {
-            $log.error("" + error);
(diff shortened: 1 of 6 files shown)


Romakita:
Hi @lwallent,

Maybe it would be nicer to not log data in log.info() but let the status in info level.

´´´typescript
request.log.info({status})
request.log.debug({status, data})
´´´
What do you think?

lwallent:
Hi @Romakita,

I followed your idea plus changed the amount of data shown for the actual request at info level.

Plus made sure that the verbose version of the request is displayed when an error is thrown, which is really what I need the most :-)

Hope this is Ok with you? I have tried it out in my own application where a massive amount of calls are made constantly, and I like the level of information now shown.

Kind Regards

Lasse :)

Romakita:
Nice :)
I'll test on my local. But I think this is good :)

PR #166, opened 2017-11-07
Feat log request

Romakita:
## Informations

Type | Status | Migration
---|---|---
Feature | Ready | No

****

## Description

@lwallent 
- I've added a configuration to filter the fields on the request. It's more flexible, I think.
- I've restore the name of the method objectToRequest because it's in the configuration documentation. And I'm remember that method is used by some developers.

I keep in mind the performance problem with the complete log. ts-log-debug is not asynchro when the stdout appender is used. But it's possible to change this behavior with a [custom appender](https://romakita.github.io/ts-log-debug/#/appenders/custom).

Maybe I'll work to add an async stdio appender (ts-log-debug).

See you
Romain

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -133,7 +133,19 @@ A call with once of this method will generate this log:
 ```
 
 The log methods is added by the [LogIncomingRequestMiddleware](api/common/mvc/logincomingrequestmiddleware.md).
-It can be overloaded with `@OverrideMiddleware`.
+
+You can configure this output from configuration:
+
+```typescript
+@ServerSettings({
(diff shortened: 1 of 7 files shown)


Romakita:
Yes. I can change the default list. What do you think about this default list:

["reqId", "method", "url", "duration"] ?

lwallent:
Well based on my original pull request the most important fields are
["method", "url"]  which will make the output 4 lines (in my setup at least)

I have no need for the "reqId" on the info level as there will be only one output.
The "duration" could be included I guess.

["method", "url", "duration"] ? :)

Issue #172, opened 2017-11-24
Mount a controller class


Ionaru:
## Information

Type | Version
---|---
Improvement | 2.14.2

****

## Description
Option to manually import controller class instead of a directory search.
```typescript
this.mount('/home', HomeController);
```

I prefer to manually import models, services and controllers without doing a directory search. Models and services can already be imported this way, but I haven't found a way for controllers. Is this possible?

Romakita:
Hi @Ionaru ,

You can do that with the mount options:

```typescript
@ServerSettings({
  "mount": {
    "/rest/": [
      "./controller/MyController1.js", 
      "./controller/MyController2.js"
    ]
  }
})
```
See you :)
Romain

Ionaru:
I'd like to do that without needing the file name, instead something like this:

```typescript
import { HomeController } from './controllers/home.controller';

@ServerSettings({
    // ...
})
export class Server extends ServerLoader {
    // ...
    public $onMountingMiddlewares(): void {
        // ...
        this.mount('/home', HomeController);
        // ...
    }
}
```

or through the settings

```typescript
import { HomeController } from './controllers/home.controller';
import { NESController } from './controllers/nes.controller';

@ServerSettings({
    // ...
    "mount": {
        "/rest/": [
            HomeController, 
            NESController
        ]
    }
})
export class Server extends ServerLoader {
    // ...
}
```

This would be very useful for me because I am using ts-node for development & debugging and in production I want to use the compiled javascript, this means I need some way of having dynamic file extensions in the code `controller.ts` vs `controller.js`, or use the direct imports.

Romakita:
@Ionaru 

Ok, I'll add your needs to the backlog :)

Romakita:
@lwallent  Can I merge this feature ?

nicojs:
Cool! I was searching for exactly this feature. Just-in-time delivery indeed! 🍰

PR #174, opened 2017-11-25
feat(server): Import manually a controller

Romakita:
## Informations

Type | Status | Migration
---|---|---
Feature | Ready | No

****

## Description

- Add ts-node support,
- Controller can be added manually to ServerSettings.mount options,
- Upgrade to TypeScript v2.5

Closes: #172

## Usage example
```
import {} from "ts-express-decorators";

@ServerSettings({
   mount: {
       "/rest": [
            "${rootDir}/controllers/Base/*.js",
            "${rootDir}/controllers/calendars/*.ts", // ts-node then fallback to .js extension
            RestCtrl // Manual import
        ],
        "/rest/v1": "${rootDir}/controllers/*.ts"
    }
})
```
> ServerLoader.mount() is also updated to support all options given in the example.

### DIFF
--- a/src/server/components/ServerLoader.ts
+++ b/src/server/components/ServerLoader.ts
@@ -363,7 +363,7 @@ export abstract class ServerLoader implements IServerLifecycle {
      * @param endpoint
      * @returns {ServerLoader}
      */
-    @Deprecated("ServerLoader.setEndpoint() is deprecated. Use ServerLoader.mount() instead of.")
+    @Deprecated("ServerLoader.setEndpoint() is deprecated. Use ServerLoader.mount() instead of. See https://goo.gl/6MPr6q.")
     /* istanbul ignore next */
     public setEndpoint(endpoint: string): ServerLoader {
 
@@ -426,24 +426,13 @@ export abstract class ServerLoader implements IServerLifecycle {
         return this;
     }
 
-    /**
-     *
-     * @param {string} file
-     * @returns {any}
-     */
-    static file(file: string): string {
-        if (!require.extensions[".ts"]) {
-            file = file.replace(".ts", ".js");
-        }
-        return file;
-    }
 
     /**
      * Add classes to the components list
      * @param classes
      * @param options
      */
-    protected addComponents(classes: any, options: { [key: string]: any } = {}): void {
+    public addComponents(classes: any | any[], options: Partial<IComponentScanned> = {}): ServerLoader {
         classes = Object
             .keys(classes)
             .map((key) => classes[key])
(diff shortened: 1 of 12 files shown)


Romakita:
@Ionaru

src/server/components/ServerLoader.ts
```diff
@@ -445,11 +475,17 @@ export abstract class ServerLoader implements IServerLifecycle {
      * @param path
      * @returns {ServerLoader}
      */
-    public mount(endpoint: string, path: string | string[]): ServerLoader {
+    public mount(endpoint: string, path: any | string | (any | string)[]): ServerLoader {
```
lwallent: Did you consider adding it as a new method:
```
public addControllers(classes:any[], endpoint: string = this._settings.endpoint, file:string=''): ServerLoader {
  this._components = (this._components || [])
                                    .concat({file, endpoint, classes});
        return this;
 }
```
The thing is that the endpoint ("endpointUrl") is already part of settings and so Controllers could be added without directly specifying it again.

Like 
```
new Server().addController([UserCtrl]).start();
```

Well, just a thought.

 Good works as allways :-) 👍

Romakita: The problem is that I want to remove endpointUrl option (this is one of the pull request) and use only the mount option with the different paths.

Components contains all classes added with mount method and scan method. Is not only for the Controlllers, you can find Services, Middlewares, Converters, etc...

So addController and mount will be have the same signature  (file parameters will be removed too. This parameter isn't used by anything). And scan() and mount() method is very redundant.

An alias can be also added, if is more comprehensible by the developer :)

What do you thinks ?

Romakita:
@lwallent Last commit add the addControllers ;)

lwallent:
@Romakita 

I like the convenience method (addControllers) and I really love the added support for ts-node. 👍 

Hours before this issue and pull request I was looking at the issue myself. I love the mount option where the controllers are dynamically picked up, however I was looking for a a way to unit test my Controllers in isolation. So I just wanted to bootstrap the server with one single controller. This PR fits great into that story.

However, it seems there is no easy way to stop the server and clean up in the testing "after" step. I guess it is assumed that the bootstraping is done once only (explains the `server.$$instance` check)

I was considering an approach where one did the following:
```
const bootstrapper = new Bootstapper();
before((done) => {
  bootstrapper
   .server(ServerMock)
   .addController(AssetsCtrl)
   .addServices(UserServiceMock)
   .launch(done);
})
after(bootstrapper.cleanup())
```
When doing TDD it is nice to have a really thin slice to work on  :-) Have you thought about this yourself? 

Sorry if this got  a little of topic :-)

Romakita:
Hi @lwallent,

No, it's good suggestion :)
bootstrap() function already exists to test run server.
```typescript

import {Server} from "../Server";
import {ExpressApplication} from "ts-express-decorators";
import {Done, bootstrap} from "ts-express-decorators/testing";
import {expect} from "chai";
import * as SuperTest from "supertest";

describe("Rest", () => {
    // bootstrap your Server to load all endpoints before run your test
    before(bootstrap(Server));

    describe("GET /rest/calendars", () => {
        it("should do something", inject([ExpressApplication, Done], (expressApplication: ExpressApplication, done: Done) => {

            SuperTest(expressApplication)
                .get("/rest/calendars")
                .expect(200)
                .end((err, response: any) => {
                    if (err) {
                        throw (err);
                    }

                    const obj = JSON.parse(response.text);

                    expect(obj).to.be.an('array');

                    done();
                });
        });
    });
});
```

So, if I take your example, We can write something like that:

```typescript
import { bootstrap, cleanupServer } from "ts-express-decorators/testing";
before(bootstrap(Server, {
    controllers: [...],
    services: [...]
}));
after(cleanupServer);
```
Do you agree with that ?

lwallent:
Hi @Romakita,

Yes, I completely agree :-) 

I have been using the bootstrap function but it is singleton instance, so the ability to cleanup the server and be able to create a new one would be really awesome! 👍 

KR
Lasse

Issue #58, opened 2017-04-18
[Decorator] @Scope on Service


RevanScript:
Hi,

Did you think about the implementation of something like Spring bean's scope (https://www.tutorialspoint.com/spring/spring_bean_scopes.htm) for services? Like the following:

```typescript
@Service
@Scope('request')
export class BlaBlaService { ...
```

For example: default scope is a singleton which is a single instance per IoC container. But it's also possible to choose **request** scope where instance exists only in the context of current HTTP request.

It would be useful to store user profile data which was obtained in the middleware for example. Of course, data from middleware can be stored within express context as a param, but it's not very comfortable to provide type safety and it isn't object oriented way :)

vincent178:
it's actually a really good topic to discuss. I like this idea.

Romakita:
Hi @RevanScript ,

Good proposal RevanScript ;). I think it's possible, but it's big improvement and I need some examples with usecase to do that. 

My question, where is injected the Service  when you use the Scope("request") decorator ?

```typescript
@Service
@Scope('request')
export class BlaBlaService {

}

export class BlablaCtrl {
       constructor(private blablaService: BlablaService) { // here ?
       }
       @Get('/') 
       getMethod(blablaService: BlablaService) { // or here ?
       
       }
}
```

See you,
Romain

RevanScript:
Here's an example how i'd like to use it:

```typescript
/*
** Service which contains user details
** UserDetailsService instance should be created on request and destroyed after (in the last middleware)
*/

@Service
@Scope('request')
export default class UserDetailsService {
    private _profile: IProfile;
    
    get profile() {
        return this._profile;    
    }
    
    set profile(profile: IProfile) {
        this._profile = profile;
    }
}

/*
** Middleware which checks auth and gets user profile
*/

@Middleware()
export default class TokenMiddleware implements IMiddleware {
    
    // Inject service into middleware
    constructor(public userDetailsService: UserDetailsService) {
        
    }

    use(@Request() request: Express.Request,
        @EndpointInfo() endpoint: Endpoint,
        @Response() response: Express.Response,
        @Next() next: Express.NextFunction) {
        
        ...
        this.userDetailsService.profile = user;
        ...
    }

}

/*
** Controller
*/

@Controller()
export class SomeController {

    // Inject service into controller
    constructor(public userDetailsService: UserDetailsService) {
        super();
    }
    
    @Get('/profile')
    @TokenRequired('profile')
    profile() {
[...]

PR #181, opened 2017-12-04
feat(mvc,di): Scope on Service, Controller and Middleware

Romakita:
## Informations

Type | Status | Migration
---|---|---
Feature | Ready | No

****

## Description
The [@Scope](docs/api/common/mvc/scope) change the behavior of a [Provider](docs/api/common/di/provider.md) like a Service, Controller or Middleware.

All providers are created one time when the server initialisation (Singleton pattern).

But [@Scope](docs/api/common/mvc/scope) decorator change that. It's mean, your Service, Controller or Middleware will be 
created each time according to the scope configuration. Actually, the only scope supported is `request`. 
So for each incoming request, a new instance of the provider will be created.

## Example

This example is available for Middleware and Controller.

```typescript
@Middleware()
@Scope("request" || ProviderScope.REQUEST) 
export class TokenMiddleware implements IMiddleware {
    // Inject service into middleware
    constructor(public userDetailsService: UserDetailsService) { }

    use(@Request() request: Express.Request,
        @EndpointInfo() endpoint: Endpoint,
        @Response() response: Express.Response,
        @Next() next: Express.NextFunction) {
        ...
        this.userDetailsService.profile = user;
        ...
    }
}
```
> In this case, the TokenMiddleware will created for each new request.

## With a Service

For the Service, it's almost the case of the previous example, but you need to keep in mind this point:

!> `@Scope` on service work only if the provider that uses it's annotated by `@Scope` decorator too.

```typescript
@Service()
@Scope("request") 
export class UserDetailsService {
   ...
}

@Middleware()
@Scope("request") // (1)
export class TokenMiddleware implements IMiddleware {
    // Inject service into middleware
    constructor(public userDetailsService: UserDetailsService) { }

    use(@Request() request: Express.Request,
        @EndpointInfo() endpoint: Endpoint,
        @Response() response: Express.Response,
        @Next() next: Express.NextFunction) {
        ...
        this.userDetailsService.profile = user;
        ...
    }
}
```
!> <sup>(1)</sup> Leaving out `@Scope("request")` on `TokenMiddleware` will give an [InjectionScopeError](api/common/di/injectionscopeerror.md).

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation


Romakita:
#58

nicojs:
What strings will be possible using `@Scope('string')`. Only 'request' and 'singleton'? If so, why not make it an enum? Or use `@RequestScoped() ` vs `@SingletonScoped()`

Romakita:
Actually the possible value is "request" or false. By default, it's always a singleton. If you use Scope() without parameters, the default value is "request".

I actually do not think about creating an enum, but I can add this, even though I think it's not mandatory.

Create two decorators will not maintainable in the future, if we add new options.

nicojs:
> Actually the possible value is "request" or false. By default, it's always a singleton. If you use Scope() without parameters, the default value is "request".

I personally would like to see it more explicit. With an enum I cannot make the mistake of misspelling `'request'`. It's also directly visible what values are possible. I also think the meaning of `@Scope(false)` and `@Scope()` is not directly clear.

Romakita:
Yes I agree ;)

Romakita:
@nicojs down :)

nicojs:
Awesome! You work fast :racehorse:

PR #183, opened 2017-12-05
feat(Example): Add Example decorator for any types

VivekGrimm:
## Informations

Fix | Ready | No

****

## Description
This is adding a new decorator for setting example for specific fields that are not strings or simple {key: string} object.
I added a new decorator so as to not change existing behaviour but feel free to change the current one instead if you think it's better, or rename the decorator if you're better at naming than me :D

## Usage example
```typescript
import {} from "ts-express-decorators";
Example(["s1", "s2"])
```

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation


src/swagger/decorators/example.ts
```diff
@@ -19,4 +19,17 @@ export function Example(name: string, description?: string) {
...
+}
+
+/**
+ * Add a example metadata of any type on the decorated element.
+ *
+ * @param {any} exampleValue
+ * @returns {(...args: any[]) => any}
+ * @decorator
+ */
+export function ExampleAnyType(exampleValue: any, ) {
```
Romakita: Maybe it's more simple to add a any type on the Example decorators :)

```typescript
export function Example(name: string | any, description?: string) {
    return (...args: any[]) => {
        let example;
        if (description) {
            example = {[name]: description};
        } else {
            example = name;
        }

        return Schema({example: example as any})(...args);
    };
}
```
I understand why you have create a new decorator, but the Example decorator has the good behavior. I had just forgot the `any` type on the signature ;)

Romakita: @VivekGrimm Thanks for your PR ;)

Romakita:
@VivekGrimm I've make the change ;). Thanks you for your PR. Example decorator can take any object as example.

Romakita:
@VivekGrimm v2.15.4 ;)

VivekGrimm:
Thanks!
Just to note though: if you put a complex object for the name parameter, and something in description, it'll try to use the complex object as a key which will likely crash.
Apart from that it will be much better like this :)

Romakita:
Haaaa.... Yes exactly :)

Issue #209, opened 2018-01-09
Enum Array Decorator for Array


lwallent:
Hi @Romakita,

Given that my beautiful EnumArray decorator didn't make it ;-) 
... allow me to ask a question

Shouldn't it be possible to extract the type from the passed Enums?

Assume we have an enum:
````
export enum SomeEnum {
    ENUM_1 = 'ENUM_1', ENUM_2 = 'ENUM_2'
}
````

and we wish to use these with Ajv - then we currently have to do the following;

````typescript
@Property({use:String})  // <---------------Needed else the type will be object
@Enum(SomeEnum.ENUM_1, SomeEnum.ENUM_2)  
someEnums?: SomeEnum[];
````

The schema: (with string)
```
    { items: { enum: [ 'ENUM_1', 'ENUM_2' ], type: 'string' },
        type: 'array' },
```

The schema: (without string)
```
    { items: { enum: [ 'ENUM_1', 'ENUM_2' ], type: 'object' },
        type: 'array' },
```

Perhaps one should look at the type passed in?

KR

Lasse

**Comment by Romakita** [comment]
Haaa, now I understand why you need a decorator to do that XD.

I think we can do that:
```typescript
@Enum(SomeEnum)  
someEnums: SomeEnum[];
```
Enum will detect if the argument is an Object (TypeScript enum) and set the type as string.
I thinks it's possible :) 

PR ?

Romain

lwallent:
Hi @Romakita 
Yes, really that is what we need. 👍 
I am not sure I can do a PR right now. I will just have to get my codebase to work with TsED 3.4.1 first :-) Ajv is not really behaving - maybe I was too early to adopt? :-)   
KR
Lasse

Romakita:
I'll work on ;)

PR #213, opened 2018-01-11
feat(jsonschema): Extends Enum decorator

Romakita:
## Informations

Type | Status | Migration
---|---|---
Feature | Ready | No

****

## Description
Now Enum decorator accept a typescript enum as parameter and build the
right jsonschema.

Closes: #209 

## Usage example

```typescript
enum SomeEnum {
    ENUM_1 = "enum1",
    ENUM_2 = "enum2"
}

 class Model {
    @Enum(SomeEnum)
    property: SomeEnum;
}
```

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation

### DIFF
--- a/src/jsonschema/decorators/enum.ts
+++ b/src/jsonschema/decorators/enum.ts
@@ -57,6 +57,34 @@ import {decoratorSchemaFactory} from "../utils/decoratorSchemaFactory";
  * }
  * ```
  *
+ * ### With Typescript Enum
+ *
+ * ```typescript
+ * enum SomeEnum {
+ *    ENUM_1 = "enum1",
+ *    ENUM_2 = "enum2"
+ * }
+ *
+ * class Model {
+ *    @Enum(SomeEnum)
+ *    property: SomeEnum;
+ * }
+ * ```
+ *
+ * Will produce:
+ *
+ * ```json
+ * {
+ *   "type": "object",
+ *   "properties": {
+ *     "property": {
+ *        "type": "string",
+ *        "enum": ["enum1", "enum2"]
+ *     }
+ *   }
+ * }
+ * ```
+ *
  * @param {string | number | boolean | {}} enumValue
  * @param enumValues
  * @returns {Function}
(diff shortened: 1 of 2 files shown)


kalu111:
Great feature!
This is what I've been waiting for :)

src/jsonschema/decorators/enum.ts
```diff
@@ -65,8 +93,15 @@ import {decoratorSchemaFactory} from "../utils/decoratorSchemaFactory";
...
  * @auto-map The data will be stored on the right place according to the type and collectionType (primitive or collection).
  */
-export function Enum(enumValue: JSONSchema6Type, ...enumValues: JSONSchema6Type[]) {
+export function Enum(enumValue: JSONSchema6Type | any, ...enumValues: JSONSchema6Type[]) {
     return decoratorSchemaFactory((schema) => {
-        schema.mapper.enum = [enumValue].concat(enumValues);
+
+        if (typeof enumValue === "object") {
+            schema.mapper.type = "string";
+            schema.mapper.enum = Object.keys(enumValue).map((key) => enumValue[key]);
```
lwallent: I am not personally using Numeric enums at the moment, but if the following are to be supported:

```typescript
enum Direction {
    Up,
    Down,
    Left,
    Right,
}
```
 shouldn't one set the type after the mapping based on the contents of `schema.mapper.enum`

Not sure, but should the above not result in:
```
"property": {
    "type": "number",
    "enum": [0, 1, 2, 3]
      }
```

Perhaps one could just say:

```typescript
schema.mapper.enum = Object.keys(enumValue).map((key) => enumValue[key]);
if (schema.mapper.enum.length > 0) { //save to assume all have the same type
  schema.mapper.type = typeof schema.mapper.enum[0];
}
```
What do you think?

Romakita: You've right, a typeof is a better solution ;)

Romakita:
@lwallent I've implemented the enum version with index, string and mixed type ;)

PR #221, opened 2018-01-23
(fix) Fixed .ts to.js scan conversion when ts-node not available

### LINKED ISSUE #220: mounting endpoints .ts file names not replaced with .js when running outside ts-node
My current server config is:

```
const rootDir = Path.resolve(__dirname);
@ServerSettings({
    rootDir,
    mount: {
        '/v1': `${rootDir}/controllers/**/**.controller.ts`
    },
    componentsScan: [
        `${rootDir}/services/**/**.service.ts`
    ],
    httpPort: process.env.PORT || 3000,
    httpsPort: false,
    acceptMimes: ['application/json'],
    swagger: {
        path: '/api-docs'
    }
})
```

when running with ts node from "./src" folder works like a charm.

but when building to dist folder and compiled to js the scanner can't find any files since it looks for .ts files.

when manually changing to .js extensions it works in prod build but fails in dev mode.

```
    componentsScan: [
        `${rootDir}/services/**/**.service.{ts,js}`
    ],
```
Doing the fix above resolves the issue, but i noticed the documentation stated it should be done automaticly when not using ts-node. 

Did I miss something ?

scopsy:
Closes #220

### DIFF
--- a/src/server/components/ServerLoader.ts
+++ b/src/server/components/ServerLoader.ts
@@ -371,8 +371,7 @@ export abstract class ServerLoader implements IServerLifecycle {
     public scan(path: string, endpoint?: string): ServerLoader {
 
         path = Path.resolve(path);
-
-        const files: string[] = require("glob").sync(path);
+        const files: string[] = require("glob").sync(ServerLoader.file(path));
 
         $log.info(`Scan files : ${path}`);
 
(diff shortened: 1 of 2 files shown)


Romakita:
Thanks @scopsy for your PR :D  
The new version will be release in a few minutes.

scopsy:
🎉  :)

PR #226, opened 2018-01-31
docs(configuration): add more info when use custom error handler.

jindev:
## Informations

Type | Status | Migration
---|---|---
Doc | Ready | No

****

## Description
add more info when create custom error handler.


Romakita:
Thanks @jindev  :)

PR #269, opened 2018-03-12
Service Interceptors

mitevs:
## Informations

Type | Status | Migration
---|---|---
Feature | In Progress | No

****

## Description
Interceptor for service methods that can access other components from tsED

## Usage example
Example to use your feature and to improve the documentation after merging your PR

PermissionInterceptor.ts
```ts
import { Interceptor, IInterceptor, IInterceptorContext } from "ts-express-decorators";

@Interceptor()
export class PermissionInterceptor implements IInterceptor {
 aroundInvoke(ctx: IInterceptorContext) {
    if (...) {
      return ctx.proceed();
    } else { 
      throw new Error('Cannot execute method');
    }
 }
}
```

UserService.ts

```ts
import { Service, Intercept } from 'ts-express-decorators';
import { PermissionInterceptor } from '../interceptors';

@Service()
export class UserService {
   @Intercept(PermissionInterceptor)
   add(userModel) {
     ...
   }
}
```

Note: Injection of services into classes is not working currently @Romakita  I get the issue that we already discussed. I have this interceptor:

screen shot 2018-03-12 at 3 00 35 pm

And this is what I get...

screen shot 2018-03-12 at 3 00 18 pm

Please check the code and suggest changes. Thanks!

## Todos
- [ ] Fix injection of services in interceptor class
- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/src/common/interceptors/decorators/interceptor.ts
+++ b/src/common/interceptors/decorators/interceptor.ts
@@ -4,11 +4,11 @@ export interface IInterceptorContext {
     target: any;
     method: string;
     args: any[];
-    proceed: <T>(err?: Error) => T;
+    proceed: <T = any>(err?: Error) => T;
 }
 
 export interface IInterceptor {
-    aroundInvoke: <T>(ctx: IInterceptorContext) => T;
+    aroundInvoke: (ctx: IInterceptorContext, options?: any) => any;
 }
 
 export function Interceptor(): Function {
--- a/src/common/interceptors/decorators/intercept.ts
+++ b/src/common/interceptors/decorators/intercept.ts
@@ -6,7 +6,7 @@ import { IInterceptor } from "./interceptor";
  * Attaches interceptor to method call and executes the before and after methods
  * @param interceptors
  */
-export function Intercept(interceptor: Type<IInterceptor>): Function {
+export function Intercept(interceptor: Type<IInterceptor>, options?: any): Function {
   return (target: any, method: string, descriptor: PropertyDescriptor) => {
     const original = descriptor.value;
 
@@ -25,7 +25,7 @@ export function Intercept(interceptor: Type<IInterceptor>): Function {
 
             throw err;
           }
-        });
+        }, options);
       }
 
       return original.apply(this, args);


Romakita:
Hi. For me, the interceptor works. Here the branch: https://github.com/Romakita/ts-express-decorators/tree/feat_interceptors
I changed nothing, except the service injected to the interceptor. Are you sure, your documentService is correctly declared as a service ?

mitevs:
@Romakita hmm, maybe I am not building it correctly? I run "npm run build", but in the index.ts I had to modify the imports from @tsed/core and @tsed/common to ./core and ./common? Do you think this might be the problem?

I don't think the services are a prob. because I can inject other services in them, and any other component in my app except the interceptors.

mitevs:
HMm, I could inject the InjectorService but when I try my services it breaks, you might be correct that my services are not ok, but I don't know what is the prob.. I don't quite understand the whole bootup process, maybe something is not correctly wired before the .load?

mitevs:
Maybe is a circular dependency because I am trying to use a service in the interceptor, but I use the interceptor on the services as well..

Romakita:
For me, all of your export is good.

Run a npm install on the Ts.ED directory.

Then, on your project, try to run theses command:
```typescript
npm link ../path/to/ts-express-decorators/src/core
npm link ../path/to/ts-express-decorators/src/common
```
Normally `@tsed/common` and `@tsed/core` will point to your local ts-express-decorators.

Romakita:
hum... I'll try with a service T.T

src/common/interceptors/decorators/interceptor.ts
```diff
@@ -0,0 +1,19 @@
...
+    proceed: <T = any>(err?: Error) => T;
+}
+
+export interface IInterceptor {
+    aroundInvoke: (ctx: IInterceptorContext, options?: any) => any;
+}
+
+export function Interceptor(): Function {
+    return (target: any): void => {
+        InjectorService.service(target);
```
Romakita: I suggest you to use this util:
```typescript
registerService(target) // instead of InjectorService.service.
```
Maybe, it'll avoid the problem with the circular dependencies. 
Note: You need to rebase your branch from the latest version (production branch).

mitevs:
Hmm, I cannot find registerService.. could you point me to the right direction?

mitevs:
Strange... when I remove the @Intercept(PermissionInterceptor) from the method in the service I can see the constructor services injected.. it only happens when I am using the interceptor on the methods that I cannot get the services injected...

Romakita:
registerService is on the latest version of Ts.ED. You need to do that:
```bash
git remote add upstream git://github.com/Romakita/ts-express-decorators.git
git stash
git rebase upstream/production
git stash pop
```
Can I have an access to your project ? Because I can't reproduce your bug actually ^^.

mitevs:
I saw your example, try to create another service and inject it in the PermissionInterceptor. For e.g. create

```ts
@Service()
export class MyService {
  sayHello() {
   return 'hello';
  }
}
```

Then instead of 
```ts
     constructor(parseService: ParseService) { 
        console.log(parseService); 
     } 
```
 do
```ts
     constructor(private myService: MyService) { 
        console.log(parseService); 
     } 
```

I will try to create separate branch on my proj and allow you to check it out. I will have to make the db mock so that you can run it without deps. Btw, I saw that if I comment the usage of the @Intercept the injection works. Seems that whenever I import the PermissionInterceptor and use it inside the @Intercept I get this issue.

mitevs:
@Romakita I've created two branches in my project that demonstrate the issue I have. I found out that the folder where I put the interceptors matters, and I got it working if I create a subfolder in my services folder 'interceptors' that contains all the interceptors and has index.ts that exports them all. I then need to import from the index.ts and attach them with the @Intercept. This way it works.

Please give me your ssh pub. key (or generate new one) so that I can give you read-access to my repo. You will have to try  both branches to see the difference. I've created the following branches...

1. no-db-working (constructor injection works)
2. no-db-notworking (constructor injection is not working)

The two branches have mocked db so that you don't have to do anything except run: "npm run dev" which will start the server in dev mode with watchers. Of course run "npm i" first to install all deps.

The first thing you should do is checkout my fork from tsED and get the production branch, rebuild and link it back to the dip-cv-rest project and then run it.

Please let me know if you have any issues. Thanks!

Romakita:
@mitevs Here my public key:

```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDDeGYYt9Heeb1Z9R4M1UDVI22rzQzYH9mnOnSQyKyvkMrNNVWOZc8iG+qLHX0ZmRz/BFxSwhWd/0r95mlRjeNr0xBZUYU3u2XgQ7QKmxTTy+1qQf9mVt6y5HMkluqg0Qi6sDVxpT0NhaJVpd3VW//whq4jFBlsNmfc+XK4tscKjT7kHyLTyYM0pqGGFytLgtfdcXA3GZ6R/+b4krsmGbQeGe03y+/1ox8x5ta5egTMyvLeWRaGZqRNFaPyp4Jso2bX0r37E3rTi4OSdOgqc5ro4tNkOBFVRs5i4EClVbDAw5yO673i9yIWw8le73xJhlCTwF+Z3+5E9K+IGBF9GZB5
```

mitevs:
@Romakita  you can check it out now, here's the repo git@bitbucket.org:code-bros/dip-cv-rest.git

mitevs:
@Romakita you will see the way I import the interceptor, but I had to export them in the interceptors/index.ts then do "export * from './interceptors'" in the services/index.ts and then I need to import the interceptor class from the services/index.ts in order for it to work. Any other way is not working, and I am assuming the issue is circular deps..

Romakita:
Yes, you have circular dependencies due to inheritance between your PermissionService and BaseService. In this case, you need to refactoring your code to avoid this circular dependency.

mitevs:
I've managed to fix it. Let me know if I should proceed with this branch for the feature. Currently I defined the interceptor decorators in my project until this goes in tsED

Romakita:
This functionality is interesting :) PR are welcome.
I've updated the feat_interceptor branch on the Ts.ED project with your last modification.

mitevs:
ok will discard this PR and continue with that branch, will submit PR from it

PR #271, opened 2018-03-14
Path Params to required for swagger validation

virus2016:
## Informations

Type | Status | Migration
---|---|---
Suggestion | Done | -

****

## Description
Path Params to required for swagger validation

## Usage example
```typescript
@Post("/")
async method(@PathParams("field") field: string) {}
```

Now has Required automatically:

```typescript
@Post("/")
async method(@Required() @PathParams("field") field: string) {}
```

### DIFF
--- a/src/common/filters/decorators/pathParams.ts
+++ b/src/common/filters/decorators/pathParams.ts
@@ -33,5 +33,5 @@ import {PathParamsFilter} from "../components/PathParamsFilter";
  * @returns {Function}
  */
 export function PathParams(expression?: string | any, useType?: any): Function {
-    return ParamRegistry.decorate(PathParamsFilter, {expression, useType});
+    return ParamRegistry.decorate(PathParamsFilter, {expression, useType, required: true});
 }
--- a/src/common/filters/registries/ParamRegistry.ts
+++ b/src/common/filters/registries/ParamRegistry.ts
@@ -135,7 +135,8 @@ export class ParamRegistry {
             target,
             useType,
             useConverter,
-            useValidation
+            useValidation,
+            required
         } = options;
 
         const param = ParamRegistry.get(target, propertyKey, parameterIndex);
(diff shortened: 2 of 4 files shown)


virus2016:
On hold, just fixing tests

virus2016:
Ready to review :-) @Romakita

src/common/filters/decorators/pathParams.ts
```diff
@@ -33,5 +33,5 @@ import {PathParamsFilter} from "../components/PathParamsFilter";
  * @returns {Function}
  */
 export function PathParams(expression?: string | any, useType?: any): Function {
-    return ParamRegistry.decorate(PathParamsFilter, {expression, useType});
+    return ParamRegistry.decorate(PathParamsFilter, {expression, useType, required: true});
```
Romakita: It's a bit strange because you don't need to set the required field to true. It's the job of the `@Required` decorator.

What about this example:
```
class MyController {
    @Get("/:id?")
    getMethod(@PathParam("id") id: string) {

   }
}
```
With your change, the PathParam will emit always a required parameter while this is not what is expected in this case.

Romakita:
I don't understand the goals of this PR. Because if you use the Required decorator, swagger documentation will be generated with the required options for the parameter. Automatize this behavior will introduce a bug when the Path parameters is optional. Express allow this path:
```
/my/path/:id?
```
See you (and thank for your proposal :) ),
Romain

virus2016:
Hi @Romakita,

It is for the swagger generation. If you have used PathParms on a endpoint, swagger expects it to be registered as required or it will not validate. 

```
/my/path/:id
```

Maybe I can change it so that if we see:

```
/my/path/:id?
```

It would not put on the require.

Many thanks,
Cal

virus2016:
We were looking to add in a mocking feature that would include faker. However, the swagger generation needs to be validated to ensure we are creating the right mocks. I think it would be extremely useful, what do you think?

Romakita:
But if you write that:
```typescript
@Post("/")
async method(@Required() @PathParams("field") field: string) {}
```
Swagger document will be rightly generated with the required value on path params no ?

virus2016:
Yeah it does. We are looking to add mocking? Do you think this would be useful?

Romakita:
Mocking ? It depend on the context. If you have an example for me ;)

Issue #252, opened 2018-02-23
SocketIO middlewares


kalu111:
Hi,
Is it possible to set middleware for socket event?

Romakita:
Hello @kalu111 ,
Is not possible actually. Socket.io don't provide a middleware mechanism. Sorry :/

See you,
Romain

kalu111:
Yes, I known that :) 
But, what about the new feature with similiar behaviour?
In my projects I very often have to check that socket is authorized before handle event listener.
What do you think to implement such feature?

It would be a event method decorator or special method in main Server class. 

Other usefull and linked feature is error handling for event listeners. After throw an error we can use @EmitError(eventName) to catch an error and send error.message to user.

kalu111:
Ping :)

Romakita:
Hi @kalu111,

We can plan new feature to handle error. Actually, I work on the new website (Localisation) for Ts.ED. 
I keep in mind your propositions ;)

See romain

Romakita:
Proposal:

```typescript
import {SocketService, SocketUse, Emit, Input, Args, SocketSession} from "@tsed/socketio";

@SocketService("/my-namespace")
export class MySocketService {
    @Input("eventName")
    @Emit("responseEventName") // or Broadcast or BroadcastOthers
    @SocketUse(MySocketMiddleware)
    async myMethod(@Args(0) userName: string, @SocketSession session: SocketSession) {

        const user = session.get("user") || {}
        user.name = userName;

        session.set("user", user);

        return user;
    }
}
```

The middleware:
```typescript
@SocketMiddleware()
export class MySocketService {
    async myMethod(@Args(0) userName: string, @SocketSession session: SocketSession) {

    }
}
```
If you have any idea. Tell me :)

kalu111:
Thanks for quick answer and proposal.
I have a few quiestions.
1. Can we inject `@Socket()` to `@SocketMiddleware()` and `@SocketMiddlewareError()` method?
2. What instance of is `@SocketErr()`?
3. If we throw an error in `@SocketMiddleware()`  or in `event handler` then `@SocketMiddlewareError()` will catch it both?
4. It is a big problem to implement a middleware that can override an `@Args()` values?.  It wolud be great to have similar feature to express Models and Converters :)

Romakita:
1. Yes, all socket decorators will be available.
2. SocketErr is required by me to inject the potential error when an error occur.
[...]

PR #278, opened 2018-03-16
feat(socket): Use middlewares on a SocketService method.

Romakita:
- Add SocketUseAfter, SocketUseBefore. Allow to add a middleware on a Socket service.
- Add SocketMiddleware, declare a new class as SocketMiddleware.

Closes: #252

### DIFF
--- a/docs/tutorials/socket-io.emd
+++ b/docs/tutorials/socket-io.emd
@@ -81,11 +81,49 @@ export class MySocketService {
 
 > @SocketService inherit from @Service decorator. That means, a SocketService can be injected to another Service, Controller or Middleware.
 
+Example:
+
+```typescript
+import * as SocketIO from "socket.io";
+import {SocketService, Nsp} from "@tsed/socketio";
+
+@SocketService()
+export class MySocketService {
+     @Nsp nsp: SocketIO.Namespace;
+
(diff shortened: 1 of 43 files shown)


virus2016:
Is there a way to use the socket in a controller to emit/broadcast?

Romakita:
Yes sir :). I just forgot this point in the document. You can expose some method from socket service then inject it into another controller or service. :)

virus2016:
Awesome! An example would be ace! I'm sure you have one quicker than I can make it :-)

Romakita:
Here an example:

SocketService: 
```typescript
@SocketService()
export class MySocketService {
     @Nsp nsp: SocketIO.Namespace;

     helloAll() {
         this.nsp.emit('hi', 'everyone!');
     }
}
```

Then inject your service in the controller:
```typescript
@Controller("/")
export class MyCtrl {
   
    constructor(private mySocketService: MySocketService) {
         
    }

    @Get('/allo')
    allo() {
         this.mySocketService.helloAll(); // BIM :)
         return "is sent";
    }
}
```

virus2016:
Brilliant! Thanks again @Romakita

Romakita:
Done :) @virus2016

PR #291, opened 2018-03-22
Added mongoose document type

troyanskiy:
## Informations

Feature | Ready | Yes

****

## Description
Added Mongoose document type for better autocompletion. I could not manage to make it with `interface`, I did it as `type`

## Usage example
The best example will be with preHook
``` typescript
@PreHook("save")
    static preSave(car: MongooseDocument<CarModel>, next) {
       if (car.model === 'Tesla') {
           car.isFast = true;
       }
       if (!car.isModified('model')) { // here I did not have a autocompletion before
           // do somthing else here
       }
       next();
    }
```

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation


troyanskiy:
I think it's pretty a lot of stuff missing in the @tsed/mongoose library.
So if you don't mind I will add some.

troyanskiy:
In fact, I have my own `tsgoose` library, and I have created it mostly because of that https://github.com/Romakita/ts-express-decorators/issues/120 and I did not see that you have some also. So I'm going to focus on yours now.

Romakita:
Your change are welcome @troyanskiy 
Thanks for your job :D

PR #319, opened 2018-04-09
fix(typescript): Fix an issue where the .d.ts file cant be found

milewski:
fix the following error message throwing on my CI:

```
node_modules/@tsed/common/lib/di/class/Providers.d.ts(2,29): error TS2307: Cannot find module '../../../core/class/Registry'.
```

this bug was introduced somewhere above version `4.12.4`


Romakita:
Thanks @milewski, my fault. I need to find a way to throw this kind of error on the compile time. Actually this error is triggered only when the package is used into another project :/.

PR #326, opened 2018-04-13
Feature/interceptors

mitevs:
Fixing binding issue for interceptor invocation.


mitevs:
@Romakita I guess we can drop this PR, the other one looks fine...

PR #327, opened 2018-04-14
fix(interceptors): Fix binding problem

Romakita:


Romakita:
@mitevs Tell me if it's the right fix ;)

mitevs:
@Romakita yep it looks good, I already tested it!

Romakita:
@mitevs  Thanks for your review. Fix will be released 4.13.5.

Issue #340, opened 2018-04-26
Circular dependency error with @Ref decorator


milewski:
the following throw an error:

```ts
@Model({})
export class Product {
    @Ref(Variant)
    variants: Variant[]
}
@Model({})
export class Variant {
    @Ref(Product)
    product: Product
}
```

```bash
[2018-04-26T16:41:01.297] [ERROR] [TSED] - TypeError: Cannot read property 'prototype' of undefined
```

Romakita:
Circular dependency is a bad pattern XD, no?

milewski:
not always.. specially when it is regarding to database instead of having a pivot table like this:

```
products ->
   variants -> ObjectID[]

variants -> 
   ....

product_variant -> 
   product_id -> ObjectID
   variant_id -> ObjectID
```

a `HasMany / BelongsTo` type of relationship might be the ideal solution in many cases... 

```
product ->
   variants -> ObjectID[]

variants -> 
  product_id -> ObjectID[]
```

Romakita:
Ok I understand, but a class cannot be used before is declaration in ES6. So the only way is to use the string name of the class like this:

```typescript
@Model({})
export class Product {
    @Ref("Variant")
    variants: Variant[]
}
@Model({})
export class Variant {
    @Ref(Product)
    product: Product
}
```
This feature isn't implemented currently. But I can patch the decorator ASAP.

Romakita:
PR ready https://github.com/Romakita/ts-express-decorators/pull/343/files

milewski:
testing this PR i get an error

```
{ INJECTION_ERROR: Service RoleService > Role not found.
    at Function.mapServices (/Users/milewski/Code/backend/node_modules/@tsed/src/common/di/services/InjectorService.ts:286:19)
    at designParamTypes.map (/Users/milewski/Code/backend/node_modules/@tsed/src/common/di/services/InjectorService.ts:257:22)
    at Array.map (<anonymous>)
......
```

and the file it is referring to import role like this

```ts
@Service()
export class RoleService {

    constructor(@Inject(Role) private model: MongooseModel<Role>) {}

}
```

Romakita:
Reopen. Error when github has closed this issue instead of #348  (my bad).

troyanskiy:
Does someone work on the issue?
I can put my eye on it since I will need that in my models.

Romakita:
I work on ;)

PR #343, opened 2018-04-30
fix(mongoose): Ref decorator support string type for circular dependency

Romakita:
#Closes: 340

### DIFF
--- a/src/mongoose/decorators/ref.ts
+++ b/src/mongoose/decorators/ref.ts
@@ -1,6 +1,6 @@
+import {PropertyMetadata, PropertyRegistry} from "@tsed/common";
 import {Store} from "@tsed/core";
 import {Schema} from "mongoose";
-import {PropertyMetadata, PropertyRegistry} from "@tsed/common";
 import {MONGOOSE_MODEL_NAME, MONGOOSE_SCHEMA} from "../constants";
 
 export type Ref<T> = T | string;
@@ -32,12 +32,14 @@ export type Ref<T> = T | string;
  * @decorator
  * @mongoose
  */
-export function Ref(type: any) {
+export function Ref(type: string | any) {
     return PropertyRegistry.decorate((propertyMetadata: PropertyMetadata) => {
-        propertyMetadata.type = type;
+        if (typeof type !== "string") {
+            propertyMetadata.type = type;
+        }
         propertyMetadata.store.merge(MONGOOSE_SCHEMA, {
             type: Schema.Types.ObjectId,
-            ref: Store.from(type).get(MONGOOSE_MODEL_NAME)
+            ref: typeof type === "string" ? type : Store.from(type).get(MONGOOSE_MODEL_NAME)
         });
     });
 }
--- a/test/units/mongoose/decorators/ref.spec.ts
+++ b/test/units/mongoose/decorators/ref.spec.ts
@@ -6,22 +6,46 @@ import {Ref} from "../../../../src/mongoose/decorators";
 import {expect} from "../../../tools";
 
 describe("@Ref()", () => {
-    class Test {
-    }
 
-    class RefTest {
-    }
+    describe("type is a class", () => {
+        class Test {
+        }
 
-    before(() => {
-        Store.from(RefTest).set(MONGOOSE_MODEL_NAME, "RefTest");
-        Ref(RefTest)(Test, "test", descriptorOf(Test, "test"));
-        this.store = Store.from(Test, "test", descriptorOf(Test, "test"));
+        class RefTest {
+        }
+
+        before(() => {
+            Store.from(RefTest).set(MONGOOSE_MODEL_NAME, "RefTest");
+            Ref(RefTest)(Test, "test", descriptorOf(Test, "test"));
+            this.store = Store.from(Test, "test", descriptorOf(Test, "test"));
+        });
+
+        it("should set metadata", () => {
+            expect(this.store.get(MONGOOSE_SCHEMA)).to.deep.eq({
+                type: Schema.Types.ObjectId,
+                ref: "RefTest"
+            });
+        });
     });
 
-    it("should set metadata", () => {
-        expect(this.store.get(MONGOOSE_SCHEMA)).to.deep.eq({
-            type: Schema.Types.ObjectId,
-            ref: "RefTest"
+    describe("type is a string", () => {
+        class Test {
+        }
+
+        class RefTest {
+        }
+
+        before(() => {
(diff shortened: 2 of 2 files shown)


milewski:
testing this PR i get an runtime error

```
{ INJECTION_ERROR: Service RoleService > Role not found.
    at Function.mapServices (/Users/milewski/Code/backend/node_modules/@tsed/src/common/di/services/InjectorService.ts:286:19)
    at designParamTypes.map (/Users/milewski/Code/backend/node_modules/@tsed/src/common/di/services/InjectorService.ts:257:22)
    at Array.map (<anonymous>)
......
```

and the file it is referring to import role like this

```ts
@Service()
export class RoleService {

    constructor(@Inject(Role) private model: MongooseModel<Role>) {}

}
```

Romakita:
@milewski Thanks for your feedback ;)

Can you send me the Role class implementation please :)

milewski:
```ts

export abstract class AbstractModel {

    @IgnoreProperty()
    _id?: Types.ObjectId

    version?: number

}

@Model({
    name: 'roles',
    schemaOptions: {
        versionKey: 'version'
    }
})
export class Role extends AbstractModel {

    @Indexed()
    @Required()
    @Unique()
    name: string

}

@Model({
    name: 'users',
    schemaOptions: {
        versionKey: 'version'
    }
})
export class User extends AbstractModel {

    @Ref(Role)
    roles: Role[]

}
```

Romakita:
@milewski I tried to reproduce your bug, but I cannot reproduce. Maybe role is missing. 
If you check the change, the fix is minor. So I think the problem come from another feature.

I just rebased the branch. Maybe the problem is fixed now ?

milewski:
i just fetched latest again rebuild and linked only the mongoose package, still getting the same error.. however i'm confuse what this PR is actually fixing? i thought you said that i can use "circular ref" like this:

```
@Model({})
export class Variant {
    @Ref("Product") //<-- this works 
    product: Product
}
@Model({})
export class Product {
    @Ref(Variant)
    variants: Variant[]
}
```

and this is working already for me

Romakita:
I'm confused too, because, this issue #340 related about a bug with the circular dependency with mongoose. ES6 doesn't allow the use of a class before its declaration. But typescript allow the use for the type annotation.

So we can't do that:

```typescript
@Model({})
export class Variant {
    @Ref(Product)  // Product at this step is undefined, and the value stored in the metadata will be undefined too and produce an error
    product: Product
}
@Model({})
export class Product {
    @Ref(Variant)
    variants: Variant[]
}
```
So this PR add the possibility to use a name instead of the class to create a mongoose reference.
```typescript
@Model({})
export class Variant {
    @Ref("Product") // Create a reference from the string name of Product class
    product: Product
}
@Model({})
export class Product {
    @Ref(Variant)
    variants: Variant[]
}
```
But if your problem is already fixed, I can postpone this PR ^^.

See you,
Romain

Romakita:
@milewski I reopened this issue, I badly referenced the PR with the issue to fix on Allow.

milewski:
yes then i dont know how it is working.. i am guessing that this '@Ref("Product")` is being passed directly to the mongoose schema, so when i do this:

```
const variant = new Variant()
const variant.product = product // the product does save only the ObjectId instead of the whole product data.. so i assume this as working.... 
variant.save()
....

**Comment by Romakita** [comment]
Yes "Product" is directly given to mongoose and produce that:

```json
{
     "type": Schema.Types.ObjectId,
     "ref": "Product"
}
```
Currently when you used the Product class as Symbol, the code get the name from the Model decorator and by default the name is "Product". But if you change the name via `@Model({name: "OProduct"})`, you need to give this name to `@Ref("OProduct")`.
Finally, It works as the same thinks, it just a manual configuration.

milewski:
OK so i guess this issue was already working from the beginning.. use "String" was the solution in this case. this PR breaks other things.. so i guess just discard it :)

Romakita:
Normally, this PR should not break anything. This is why it's strange when you said me, I've a problem with the injector. I think, if you install the latest version of Ts.ED, normally (sick T.T), you should have the same problem (So the problem would come from elsewhere).
Can you try the latest version of Ts.ED, and tell me if you have the same problem :) ?

milewski:
i do not face any problem with the version `4.17.1` all my test are green, 

And the way im testing only the mongoose package is (just in case it depends on something else im not aware of)

1. clone this repo
2. checkout branch: `fix_340_ref_mongoose`
3. npm install && npm build
4. cd dist/packages/mongoose
5. npm link

go to my project and run `npm link @tsed/mongoose` but once i do it... i cant start my project anymore due to the error i pointed out

Romakita:
haaa ok, npm link directly on ts-express-decorators project failed in my case because you have subpackage like "@tsed/mongoose".
My process with npm link is different. For example, if I have that:
```typescript
|- myproject
|- ts-express-decorators
```
I use these commands:
```typescripts
cd myproject
npm link ../ts-express-decorators/src/core
npm link ../ts-express-decorators/src/common
npm link ../ts-express-decorators/src/mongoose
npm link ../ts-express-decorators/src/swagger
```
I need to link each packages. If you link only one (like just mongoose). Node will resolve two tsed module, one with from your local and another one from node_modules. In this case, the InjectorRegistry will doesn't know your model declared with the linked package because it don't share the same module (Node and the resolving module).

Try my process, I think your project will work with that ;)

milewski:
linking that way i get this errors:

```
G:\Code\node_modules\ts-node\src\index.ts:330
throw new TSError(formatDiagnostics(diagnosticList, cwd, ts, lineOffset))
^
TSError: Γ¿» Unable to compile TypeScript
    ..\ts-express-decorators\src\ajv\services\AjvService.ts (7,8): Cannot find module '@tsed/common'. (2307)
    ..\ts-express-decorators\src\ajv\services\AjvService.ts (8,28): Cannot find module '@tsed/core'. (2307)
    ..\ts-express-decorators\src\ajv\services\AjvService.ts (11,26): Cannot find module 'ts-httpexceptions'. (2307)
    ..\ts-express-decorators\src\ajv\services\AjvService.ts (12,20): Cannot find module 'ts-log-debug'. (2307)
```
although the packages does exists into the  node_modules folder

image

milewski:
but yeah linking all packages instead of just mongoose doesnt throw that error anymore..

``` 
cd dist/packages ** npm link **/
cd ~/my-project && 
     npm link @tsed/ajv 
     ..... all the other packages im using
```

and all my tests still pass.. so i take back what i said: this pr doesnt break anything :)

Romakita:
Ok go to release this fix :)

Issue #348, opened 2018-05-03
@Allow() decorator not working


milewski:
Have just tried this out and looks like this decorator is not working:

```ts
export class userModel {

    @Required() // <-- have also tried passing '' in here
    @Allow('') // <-- preferably setting this to null would be better than empty string
    email: string

    @Required()
    @Allow('') 
    password: string

}
```

```ts
const user = await userModel.create({ password: null, email: null }) // doesn't work
const user = await userModel.create({ password: '', email: '' }) // doesn't work
const user = await userModel.create({ password: 'abc', email: 'abc' }) // works
```

Error

```
users validation failed: password: Path `password` is required., email: Path `email` is required.
```

optimally i think setting `@Allow(null)` is better than an empty string

Romakita:
Ok I found the problem. After the fix, you can use allow as following:

```typescript
export class userModel {

    @Required() 
    @Allow('', null) // allow accept a list of value (null, '', undefined)
    email: string

    @Required()
    @Allow('', null) 
    password: string
}
```

Romakita:
Fixed :)

Romakita:
Ok. Can you send me the code of the `create` method. 
I tested with that:

```typescript
export class UserModel {

    @Required()
    @Allow(null)
    email: string

    @Required()
    @Allow(null)
    password: string
}

ConverterService.deserialize({email: null, password: null}, UserModel);
```
As you had mentioned in your first example. Now the problem is different because, the error, now,  come from mongoose and Allow has any effect with Mongoose validator.

Romakita:
The problem is you have a many validator in your configuration. ConverterService validate the required value (with allow), then you have the AJV and to finish you have the mongoose. Conciliate all validators it's a hard work XD.

Romakita:
I'll check if allow is possible with Mongoose.

milewski:
the create method comes from mongoose: http://mongoosejs.com/docs/api.html#create_create

One option could be defining the required prop as a function and do some checking to allow the value or not... something like this: 
```
email: {
    type: String,
    required: function() {
        return doSomeCheckAgaintsTheGivenValue(value);
    }
}
```

PR #352, opened 2018-05-05
fix(mvc): Fix allow when Required and Allow are used at the same time

Romakita:
Closes: #348

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "@types/socket.io": "^1.4.31",
     "@types/superagent": "^3.5.3",
     "@types/supertest": "^2.0.0",
-    "@types/swagger-schema-official": "^2.0.7",
+    "@types/swagger-schema-official": "^2.0.9",
     "ajv": "^5.2.4",
     "body-parser": "^1.15.1",
     "chai": "^4.1.0",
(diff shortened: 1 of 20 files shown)


Romakita:
Now this fix will work @milewski ;)
Thanks for your help (with the mongoose documentation)

milewski:
still not working :( got the same error.. and have found some other "blockers" related to this issue on the way... 

```ts
export class UserModel {

    @Indexed()
    @Required()
    @Email()
    @Unique()
    @Allow(null)
    email: string

    @Required()
    @Minimum(6)
    @Allow(null)
    password: string

}

/// on my tests

const userData = {
    email: null,
    password: null
}

await this.agent.post('/v1/user/create').send(userData)


/// on the controller

@Post('/user/create')
@Status(201)
async createUser(@BodyParams() userData: UserModel) { 
     await this.userService.create(userData)
}

// error 
@Email() fail because null is not a email
@Minimum(6) fail because null is not minimum 6

email:string, password: string, fails because null is not a string

```

and disabling these checks and letting the request hit the user service still gets

```users validation failed: password: Path `password` is required., email: Path `email` is required.```

Romakita:
@milewski Ok I'll try your example and fix it. Tell me If you find others problems :)

Romakita:
Work fine now. Just one point, in your example replace `Minimum` by `MinLength` ;).

See you

PR #357, opened 2018-05-06
Code style 2

troyanskiy:
## Informations

Type | Status | Migration
---|---|---
Fix | Ready | No

****

## Description
Implementation of new rules and code refactoring.

Rules added

- prefer-const
- newline-before-return
- no-angle-bracket-type-assertion
- only-arrow-functions
- arrow-return-shorthand
- object-literal-shorthand
- array-type
- variable-name


troyanskiy:
@Romakita Could you please have a look at my commits... Nothing dangerous, just code styling and rules on tslint.

Issue #367, opened 2018-05-14
Ajv Validation Error Response in JSON


saasrack:
## Informations

Type | Version
---|---
Question | 4.14

****

## Description
I have implemented the `@tsed/ajv` validation package. However, I have hard time to figure out how to response validation errors in JSON format to make it readable to the consuming party. It seems that the `ParseExpressionError` class containing `buildMessage` function is only able to return a pre-set string with some information regarding the validation error.

```javascript
static buildMessage(name: string, expression: string | RegExp | undefined, message?: string) {
    name = name.toLowerCase().replace(/parse|params|filter/gi, "");
    return `Bad request on parameter "request.${name}${expression ? "." + expression : ""}".\n${message}`.trim();
}
```

Is there a way to passthrough the Ajv validation error object or override the response?

Thanks.

Romakita:
Hi @saasrack,

Currently not. But we can work on this, to passthrough the AJV validation error in the response.

I'll work on ASAP.
See you,
Romain

Romakita:
Sorry I've completely forgot to finish this feature. Now is released ;) v4.20.0

milewski:
@Romakita can you show an example on how to use this feature? im also in need of something similar

PR #368, opened 2018-05-16
feat: Create custom error with headers and errors field.

Romakita:
Introduce a new interface IResponseError which allow to create custom
error based on BadRequest, NotFound, etc... When a custom error implement
this interface and fill the headers or errors fields, these information will
be forwarded to the GlobalErrorHandler and will be used to generate the
 response headers (errors generate the header `errors = "[]"` formatted in JSON).

- fix swagger documentation with `@Any` decorator.

Closes: #367


Romakita:
@kalu111 effectively, the error management on Socket isn't implemented. It should do the same things. 

Customized name for errors field? why not. It's just a key configuration in server settings.

Romakita:
@kalu111 Socket doesn't define a format to send error. This is why I haven't implemented a global error handler for the socket. With the UseSocketAfter decorator on your class, you can handle all error and implement your own error handler strategy, right ?

kalu111:
@Romakita Yes, You have right about socket error handling.
In my applications I have implemented error handling by using `@SocketMiddlewareError()`.

Thanks,
Bye.

Romakita:
But we can use your Middleware Error as example :)

PR #439, opened 2018-08-28
Removed render decorator

joost:
Copy-pasting tutorial broke with `Error: No default engine was specified and no extension was provided.`.


joost:
Hmm.. seems that the example shows more, like `@Authenticated`. Seems intentional?

joost:
I was just copy-pasting through the documentation and didn't look much further :)
My suggestion would be to remove them since those decorators are not explained in this part of the documentation.

Issue #445, opened 2018-09-04
Swagger-ui doesn't work with IE 11


Tlahey:
## Informations

Type | Version
---|---
Bug | 4.30.1

****

## Description
Error when I go on the swagger with IE 11.

I found the issue on `views/index.ejs` : 
 - Arrows functions doesn't work on IE 11
 - Object `init` must be set to a function
 - Use `window.location.origin` insted of `window.origin`
 - `Object.assign` [doesn't exist on IE 11, so we need to put a polyfill](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Object/assign#Proth%C3%A8se_d'%C3%A9mulation_(polyfill)) 

That's my code with the remplacements : 

```ejs

[...]

<script src="./swagger-ui-bundle.js"></script>
<script src="./swagger-ui-standalone-preset.js"></script>
<script>
  if (typeof Object.assign != 'function') {
    // Must be writable: true, enumerable: false, configurable: true
    Object.defineProperty(Object, "assign", {
      value: function assign(target, varArgs) { // .length of function is 2
        'use strict';
        if (target == null) { // TypeError if undefined or null
          throw new TypeError('Cannot convert undefined or null to object');
        }

        var to = Object(target);

        for (var index = 1; index < arguments.length; index++) {
          var nextSource = arguments[index];

          if (nextSource != null) { // Skip over if undefined or null
            for (var nextKey in nextSource) {
              // Avoid bugs when hasOwnProperty is shadowed
              if (Object.prototype.hasOwnProperty.call(nextSource, nextKey)) {
                to[nextKey] = nextSource[nextKey];
              }
            }
          }
        }
        return to;
      },
      writable: true,
      configurable: true
    });
  }

[...]

  const origin = window.location.origin;

[...] 

  .map(function(o) {
    if (!o.url.match(/^https?:/)) {
      const url = origin + o.url;
      return {
        name: o.name,
        url: url,
        selected: url === currentUrl
      }
    }
    return o;
  })
    .sort(function(a, b){
      return a.selected ? -1 : 1
    });

[...]

      layout: "StandaloneLayout",
      validatorUrl: window.location.origin.match('0.0.0.0') || window.location.origin.match('localhost') ? null : undefined

[...]

    init : function() {
      this.ui = SwaggerUIBundle(this.config);
[...]

```

Thanks

Tlahey:
thank you for the correction but in corretif it is missing : 
[...]

PR #448, opened 2018-09-05
fix(swagger): Add IE 11 support for swagger-ui

Romakita:
Closes: #445


Tlahey:
Hello @Romakita,

Merge need an another fix, see last comment #445 

Thanks.

PR #478, opened 2018-12-04
fix(di): get symbol description instead of its constructor

ArtemDerevnjuk:
## Informations

Type | Migration
---|---
Fix | No

****

## Description
Provider uses constructor symbol instead of its description. This results in that `InjectorService` creates few instances on one provider. 

See following example:

```typescript
import {InjectorService} from "@tsed/common";

export interface DummyService {
   noop(): void;
}

export const DummyService = Symbol("DummyService");

export class DummyServiceImp {}

registerProvider({provide: DummyService, userClass: DummyServiceImp});

export interface DummyService2 {
   noop(): void;
}

export const DummyService2 = Symbol("DummyService2");

export class DummyService2Imp {}

registerProvider({provide: DummyService2, userClass: DummyService2Imp});
```
### Expected behavior: 

If I try to call `InjectorService.load()` method it builds 2 instances for following class DummyService2Imp and DummyServiceImp.

### Actual behavior:

It builds instances more than enough.

### DIFF
--- a/packages/common/src/di/class/Provider.ts
+++ b/packages/common/src/di/class/Provider.ts
@@ -1,4 +1,4 @@
-import {getClass, nameOf, NotEnumerable, Store, Type} from "@tsed/core";
+import {getClass, getClassOrSymbol, nameOf, NotEnumerable, RegistryKey, Store, Type} from "@tsed/core";
 import {ProviderScope} from "../interfaces";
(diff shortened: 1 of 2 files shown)


Romakita:
Thanks @ArtemDerevnjuk for your work ;)

PR #497, opened 2018-12-21
Feat/test context api

Romakita:
## Informations

Type | Migration
---|---
Feature | Yes

****

## Description
Improve TestContext API


Romakita:
@nicojs Just a small PR to improve the TestContext API

packages/testing/src/bootstrap.ts
```diff
@@ -1,24 +1,15 @@
...
-    instance.startServers = () => Promise.resolve();
 
-    // used by inject method
-    TestContext.injector = instance.injector;
-
-    return instance.start();
-  };
+/* istanbul ignore next */
+export function bootstrap(server: any, ...args: any[]): () => Promise<void> {
+  return TestContext.bootstrap(server, {arguments: args});
```
nicojs: Maybe warn users with a deprecation warning?

packages/testing/src/inject.ts
```diff
@@ -9,14 +8,11 @@ import {TestContext} from "./TestContext";
...
-    const injector = TestContext.injector || loadInjector();
 
-    const args = targets.map(target => (injector.has(target) ? injector.get(target) : injector.invoke(target)));
-    const result = await func(...args);
-
-    return result;
-  };
+/* istanbul ignore next */
+export function inject<T>(targets: any[], func: (...args: any[]) => Promise<T> | T): () => Promise<T> {
+  return TestContext.inject(targets, func);
```
nicojs: Same here

packages/testing/src/invoke.ts
```diff
@@ -1,13 +1,12 @@
...
 export function invoke(target: Type<any>, providers: {provide: any | symbol; use: any}[]) {
-  const injector = loadInjector();
-  const locals = new Map();
-
-  providers.forEach(p => {
-    locals.set(p.provide, p.use);
-  });
-
-  return injector.invoke(target, locals);
+  return TestContext.invoke(target, providers);
```
nicojs: Same here

packages/testing/src/bootstrap.ts
Romakita: Hum maybe is not right moment to deprecated theses function. I'll deprecate theses function when the TestContext will be completly adopted / stable ;)
Thanks nicojs :)

Romakita:
I removed the deprecation annotation on bootstrap. Now it's fine :). We can merge ?

PR #508, opened 2019-01-19
Feat/virtual schema mongoose

Romakita:


Romakita:
@alex-dev Ok the problem with the CI is related to a specific bug between node v11 and istanbul. I'll investigate on and fix that

PR #526, opened 2019-03-26
fix(di): Remove constant serialization

RevanScript:
## Informations

Type | Breaking change
---|---
Fix | Yes/No

****

## Description
Serialization doesn't allow to pass functions to settings


RevanScript:
ServerSettingsService tests failed. Will check when have time.

Romakita:
@RevanScript I fixed the problem ;). Because we apply directly an Object.freeze on the obj, it freeze the object in ServerSettingsService also. I added a new  deepClone function in `@tsed/core` package to clone correctly an object with preserving function/classe.
You configuration should work correctly in case of `@tsed/graphql` ^^

See you

Romakita:
v5.5.0

RevanScript:
Awesome! Will test it in the evening. Thanks

PR #527, opened 2019-03-27
feat(graphql): Add instance and schema to servers map

RevanScript:
## Informations

Type | Breaking change
---|---
Feature | Yes

****

## Description
GraphQLService allows to get ApolloServer instance only and it's impossible to get a generated Schema. I changed `_instances` map to object that contains 2 properties currently:

* instance
* schema

Is this fine? Because I'm not sure if it's the best option and maybe there's a better way to do it :)

## Todos

- [x] Tests
- [x] Coverage
- [x] Example

### DIFF
--- a/packages/graphql/src/services/GraphQLService.ts
+++ b/packages/graphql/src/services/GraphQLService.ts
@@ -27,7 +27,7 @@ export class GraphQLService {
 
   /**
    *
-   * @returns {Promise<"mongoose".Connection>}
+   * @returns {Promise<ApolloServer>}
    */
   async createServer(id: string, settings: IGraphQLSettings): Promise<any> {
     const {
@@ -92,7 +92,7 @@ export class GraphQLService {
 
   /**
    * Get an instance of ApolloServer from his id
-   * @returns {"mongoose".Connection}
+   * @returns ApolloServer
    */
   get(id: string = "default"): ApolloServer | undefined {
     return this._servers.get(id)!.instance;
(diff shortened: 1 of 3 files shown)


packages/graphql/src/services/GraphQLService.ts
```diff
@@ -89,8 +93,8 @@ export class GraphQLService {
    * Get an instance of ApolloServer from his id
    * @returns {"mongoose".Connection}
    */
-  get(id: string = "default"): ApolloServer | undefined {
-    return this._instances.get(id);
+  get(id: string = "default"): IGraphQLServer | undefined {
```
Romakita: It could be better to create a seperated method like `getSchema(id: string = default)` to retrieve the schema.

The `get(id: string = "default")` should keep the same return type `ApolloServer` to avoid breaking change. 
Then in major release 6.0.0 (this release is in progress with #521 ) we will change the method by `get(id: string = "default"): IGraphQLServer | undefined`

;)

Romakita: Nice job otherwise ;)

RevanScript: Yeah, I was thinking about `getSchema` method to avoid breaking change, but didn't like the idea with passing server name twice in case when I need both instances :)

I'll commit `getSchema` today. Should I keep `IGraphQLServer` in this commit btw? (If we are going to use it in 6.0.0). Or better to remove it for now?

RevanScript: I think it can be merged now :)

Romakita: Thanks @RevanScript I'll fix the coverage and release the package ;)

RevanScript: Ok, thanks!

PR #543, opened 2019-05-07
Feat/seq

OskarLebuda:
## Informations

Type | Breaking change
---|---
Feature|Yes

****

## Description
A new package for the system logs. Seq is centralized structured logs which provides the gui where we can view all application logs without any terminal/console

## Usage example
```typescript
import {ServerLoader, ServerSettings} from "@tsed/common";
import "@tsed/seq"; // import seq Ts.ED module

@ServerSettings({
  rootDir: __dirname,
  seq: {
    url: "http://localhost:5341"
  }
})
export class Server extends ServerLoader {

}
```

`Controller.ts`
```typescript
import {Controller, Get, Post} from "@tsed/common";
import {$log} from "ts-log-debug";

@Controller('/calendars')
export class Calendar {
    
    @Get('/:id')
    async getCalendar(@QueryParam() id: string): Promise<CalendarModel> {
      $log.info(id);
    }
   
}
```
## Todos

- [ ] Tests


packages/seq/src/SeqModule.ts
```diff
@@ -0,0 +1,21 @@
+import {Inject, Service} from "@tsed/di";
+import {$log} from "ts-log-debug";
+import {AfterRoutesInit, ServerSettingsService} from "@tsed/common";
+import {ISeqSettings} from "./interfaces/ISeqSettings";
+
+@Service()
+export class SeqModule implements AfterRoutesInit {
+  private config: ISeqSettings;
```
Romakita: You can use @Constant decorator instead of service Inject on contructor.

```typescript
export class SeqModule implements AfterRoutesInit {
  @Constant('seq', {})   // second parameter is the default value of the property seq 
  private config: ISeqSettings;
```

packages/seq/src/SeqModule.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  constructor(@Inject() private serverSettingsService: ServerSettingsService) {
```
Romakita: And remove the constructor

Romakita:
@OskarLebuda Nice job ;). Just small change before merging this PR ;)

OskarLebuda:
@Romakita It's done! 👍  Thanks for your review

PR #549, opened 2019-05-10
Fixed #546

### LINKED ISSUE #546: Virtual Refs not possible to pass aditional configurations
## Information

- **Version:** 5.13.0
- **Type:** Issue

## Description
I would like to pass aditional configurations to virtual refs, such as "justOne" and "localField", this isn't possible right now but is very easy doable.

## Example
In my case, I have a class which contains a field "account_id", which is the reference to the other Model. Also, I have a field called "account" which is my virtual reference, the thing is, I don't always need to populate that field but I always want the referenced id and in my case, my "account" field should be just an object when populated, not an array, and this is possible by passing an option "justOne: true" to the virtual reference configuration.

Below is my class that contains my virtual reference:

```
export class IAccountData {
    public _id?: ObjectId;

    @Required()
    public type?: String;

    @Required()
    public key?: String;

    @Required()
    @Default({})
    public value?: Object;

    @Ref(IAccount)
    @Required()
    public account_id?: Ref<IAccount>;

    @VirtualRef('Account', '_id')
    @PropertyName('account_id')
    public account?: VirtualRef<IAccount>;
}
```

You can see that I have to set the @PropertyName('account_id') to make it work, since it's the only way I found to override the "localField" property. But I can't set the "justOne" property which is more important.

To fix that, we can just do something like:

(virtualRef.ts)
```
export function VirtualRef(type: string, foreignField: string, justOne: Boolean = false) {
  return PropertyRegistry.decorate((propertyMetadata: PropertyMetadata) => {
    propertyMetadata.store.merge(MONGOOSE_SCHEMA, {
      ref: type,
      localField: propertyMetadata.name,
      foreignField,
      justOne
    });
  });
}
```

Or instead of passing more parameters, accept just one parameter of type Object which will contain all the configurations

ghmendonca:
## Informations

Type | Breaking change
---|---
Fix | No

****

## Description
Fixed issue #546  

## Usage example
Now you can use both ways, passing an options object or two string parameters:

```
export class IAccountData {
    public _id?: ObjectId;

    @Required()
    public type?: String;

    @Required()
    public key?: String;

    @Required()
    @Default({})
    public value?: Object;

    @Ref(IAccount)
    @Required()
    public account_id?: Ref<IAccount>;

    @VirtualRef({type: 'Account', foreignField: '_id', localField: 'account_id', justOne: true})
    public account?: VirtualRef<IAccount>;
}
```

Or

```
export class IAccountData {
    public _id?: ObjectId;

    @Required()
    public type?: String;

    @Required()
    public key?: String;

    @Required()
    @Default({})
    public value?: Object;

    @Ref(IAccount)
    @Required()
    public account_id?: Ref<IAccount>;

    @VirtualRef('Account', '_id')
    public account?: VirtualRef<IAccount>;
}
```
Note that if you ommit "localField" from the options object, it will take from the property metadata name.

I didn't create a test case for it, but I just tested in my project and it works like a charm.

### DIFF
--- a/packages/mongoose/src/interfaces/MongooseVirtualRefOptions.ts
+++ b/packages/mongoose/src/interfaces/MongooseVirtualRefOptions.ts
@@ -1,0 +1,7 @@
+export interface MongooseVirtualRefOptions {
+  type: string;
+  foreignField: string;
+  localField?: string;
+  justOne?: boolean;
+  options?: object;
+}
--- a/packages/mongoose/src/decorators/virtualRef.ts
+++ b/packages/mongoose/src/decorators/virtualRef.ts
@@ -1,5 +1,6 @@
 import {PropertyMetadata, PropertyRegistry} from "@tsed/common";
 import {MONGOOSE_SCHEMA} from "../constants";
+import {MongooseVirtualRefOptions} from "../interfaces/MongooseVirtualRefOptions";
 
 export type VirtualRef<T> = T | null;
 export type VirtualRefs<T> = T[];
@@ -38,12 +39,26 @@ export type VirtualRefs<T> = T[];
  * @decorator
  * @mongoose
  */
-export function VirtualRef(type: string, foreignField: string) {
+
+export function VirtualRef(type: string, foreignField: string): Function;
+export function VirtualRef(options: MongooseVirtualRefOptions): Function;
+
+export function VirtualRef(options: string | MongooseVirtualRefOptions, foreignField?: string): Function {
   return PropertyRegistry.decorate((propertyMetadata: PropertyMetadata) => {
-    propertyMetadata.store.merge(MONGOOSE_SCHEMA, {
-      ref: type,
-      localField: propertyMetadata.name,
-      foreignField
-    });
+    if (typeof options === "object") {
+      propertyMetadata.store.merge(MONGOOSE_SCHEMA, {
+        ref: options.type,
+        localField: options.localField || propertyMetadata.name,
+        foreignField: options.foreignField,
+        justOne: options.justOne || false,
+        options: options.options
+      });
+    } else {
+      propertyMetadata.store.merge(MONGOOSE_SCHEMA, {
+        ref: options,
+        localField: propertyMetadata.name,
+        foreignField
+      });
+    }
   });
 }
--- a/packages/mongoose/test/decorators/virtualRef.spec.ts
+++ b/packages/mongoose/test/decorators/virtualRef.spec.ts
@@ -1,23 +1,75 @@
-import {Store} from "@tsed/core";
-import {descriptorOf} from "@tsed/core";
+import {descriptorOf, Store} from "@tsed/core";
 import {MONGOOSE_SCHEMA} from "../../src/constants";
 import {VirtualRef} from "../../src/decorators";
-import {expect} from "chai";
 
 describe("@VirtualRef()", () => {
-  class Test {}
-  class RefTest {}
+  describe("when type and foreign value are given", () => {
+    it("should set metadata", () => {
+      // GIVEN
+      class Test {
+      }
 
-  before(() => {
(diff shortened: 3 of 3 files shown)


packages/mongoose/src/interfaces/MongooseVirtualRefOptions.ts
```diff
@@ -0,0 +1,7 @@
+export interface MongooseVirtualRefOptions {
+  type: string;
+  foreignField: string;
+  localField?: string;
+  justOne?: boolean;
+  options?: object;
```
Romakita: I think it's any intead of object ? 
options is a possible field used by virtual ref ?

Romakita:
@ghmendonca Thanks for your PR, just minor change and you forgot unit test ;)
https://github.com/TypedProject/ts-express-decorators/blob/master/packages/mongoose/test/decorators/virtualRef.spec.ts

to help you, I wrote the unit test (with the new format):

```typescript
import {descriptorOf, Store} from "@tsed/core";
import {MONGOOSE_SCHEMA} from "../../src/constants";
import {VirtualRef} from "../../src/decorators";

describe("@VirtualRef()", () => {
  describe("when type and foreign value are given", () => {
    it("should set metadata", () => {
      // GIVEN
      class Test {
      }

      const store = Store.from(Test, "test", descriptorOf(Test, "test"));
      // WHEN
      VirtualRef("RefTest", "foreign")(Test, "test", descriptorOf(Test, "test"));

      // THEN
      store.get(MONGOOSE_SCHEMA).should.deep.eq({
        ref: "RefTest",
        foreignField: "foreign",
        localField: "test"
      });
    });
  });

  describe("when options is given with minimal fields", () => {
    it("should set metadata", () => {
      // GIVEN
      class Test {
      }

      const store = Store.from(Test, "test", descriptorOf(Test, "test"));
      const options = {type: "RefTest", foreignField: "foreign"};
      // WHEN
      VirtualRef(options)(Test, "test", descriptorOf(Test, "test"));

      // THEN
      store.get(MONGOOSE_SCHEMA).should.deep.eq({
        ref: "RefTest",
        localField: "test",
        foreignField: "foreign",
        justOne: false,
        options: undefined
      });

    });
  });

  describe("when options is given with all fields", () => {
    it("should set metadata", () => {
      // GIVEN
      class Test {
      }

      const store = Store.from(Test, "test", descriptorOf(Test, "test"));
      const options = {
        type: "RefTest",
        foreignField: "foreign",
        localField: "test_2",
        justOne: true,
        options: {}
      };
      // WHEN
      VirtualRef(options)(Test, "test", descriptorOf(Test, "test"));

      // THEN
      store.get(MONGOOSE_SCHEMA).should.deep.eq({
        ref: "RefTest",
        localField: "test_2",
        foreignField: "foreign",
        justOne: true,
        options: {}
      });
    });
  });
});
```

Good job,
See you,
Romain

packages/mongoose/src/interfaces/MongooseVirtualRefOptions.ts
ghmendonca: For what I read in mongoose documentation about virtual refs, it expects an object and not any (https://mongoosejs.com/docs/api.html#query_Query-setOptions). Do you still want me to change or let's stick with what mongoose expects?

ghmendonca:
Thanks for writing the unit test! I will commit it. Now, we just need to align the change you requested 😄

packages/mongoose/src/interfaces/MongooseVirtualRefOptions.ts
ghmendonca: And also, virtual refs accepts options: https://mongoosejs.com/docs/populate.html#populate-virtuals

Romakita: Not necessary. It's fine for me :)

Romakita:
Thanks @ghmendonca ;)

ghmendonca:
Awesome! Thanks @Romakita

PR #569, opened 2019-06-13
Fix typo so the documentation can be build correctly

Szauka:
Type | Breaking change
---|---
Fix/Doc| No

****

## Description
image


Romakita:
You're change is just the bracket  ?

Szauka:
@Romakita sorry, I should have explained better this. 
The issues is with the file name: 
`marge-params-1.ts`  before, notice the **a** from `marge`
`merge-params-1.ts`  change made on the pull request
But the only way create the pull request was to make a change on the file content itself as the renaming of the file isn't piked up as a change.

Romakita:
Ha ok good :)

PR #586, opened 2019-07-08
Fix broken issue page links

kennanseno:
## Informations

Type | Breaking change
---|---
Doc | No


Romakita:
Thanks @kennanseno ;)

Issue #579, opened 2019-06-25
TypeORM predefined connection object


marghidanu:
## Information

- **Version:** 5.x
- **Type:** Issue

## Description

Hello, 

We are trying to create a test with **supertest** that uses two instances of ServerLoader apps both using the same database. The problem is that TypeORM manages connections is some sort of a Singleton object thus causing the application to fail silently. It would be nice if TypeORMService would check if the connection already exists using `getConnection(name)` from TypeORM and if the connection already exists it should reuse it.

Romakita:
Hello @marghidanu,
Can you help me by creating a small PR with the required change (or copy paste here the code) ? I'll doing unit test for you ^^
Here the TypeORMService code https://github.com/TypedProject/ts-express-decorators/blob/master/packages/typeorm/src/services/TypeORMService.ts
See you
Romain

PR #590, opened 2019-07-11
fix: TypeORMService (createConnection -> getConnectionManager)

chr33s:
## Informations

Type | Fix

****

## Description
should fix: #579

### DIFF
--- a/packages/typeorm/src/services/TypeORMService.ts
+++ b/packages/typeorm/src/services/TypeORMService.ts
@@ -1,6 +1,6 @@
 import {Service} from "@tsed/common";
 import {$log} from "ts-log-debug";
-import {Connection, ConnectionOptions, createConnection} from "typeorm";
+import {Connection, ConnectionOptions, getConnectionManager} from "typeorm";
 
 @Service()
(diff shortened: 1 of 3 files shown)


packages/typeorm/src/services/TypeORMService.ts
```diff
@@ -39,23 +42,6 @@ export class TypeORMService {
     }
   }
 
-  /**
-   *
-   * @returns {"mongoose".Connection}
-   */
-  get(id: string = "default"): Connection | undefined {
```
Romakita: Why have you removed the get and has method ? You have to keep this method, otherwise you'll create a breaking change :)

Romakita:
Hello @chr33s,
Thanks for your PR :)
You've forgot to write unit test :)

chr33s:
my bad, I reverted the breaking API changes. Didn't include tests due to 
> [I'll doing unit test for you ^^](https://github.com/TypedProject/ts-express-decorators/issues/579#issuecomment-506218678) ;), easy enough tho, will add some when I have a free moment.

PR #606, opened 2019-07-31
Throw bad request error when property convert failed on Converter Service

NitzanHardonVonage:
ConvertService throws internal server error when it fails to convert a decorated property.
The expected behavior is to throw a bad request error.

example: 
When expecting to receive a body param of the following schema:
```
class User {
  @Required()
  id: number;
}
```

And the HTTP body request contains  
```
{ 
  id="sdn12nsf"
} 
```
 
The expected HTTP response is:
```
{
"status": 400,
"body": "Error for id with value sdn12nsf \nCast error. Expression value is not a number."
}
```

The current HTTP response is:
```
{
    "status": 500,
    "message": "Internal Server Error - Conversion failed for class \"User\" with object => {\"id\":\"sdn12nsf\"}.\nnewMessage is not defined"
}
```


Romakita:
@NitzanHardonVonage Thanks for you PR :)

PR #630, opened 2019-08-26
Stop outputting the settings

pirouet:
## Informations

Type | Breaking change
---|---
Fix |No

****

## Description
The idea of this commit is to remove the command to log out each configured setting. The reason for this is to improve security since TypeOrm - for example - settings are displayed in the logs; this includes passwords and username.

This should not be outputted since it provides an attacker with a way to identify the connections needed. These connection(s) could be access to a critical Production system therefore it should be up to the developer to decide if they wish to write data to the logs.

If needed, the settings object can be accessed from the Server object returned from `ServerLoader.bootstrap()` thus ensuring any decision is thought

Tests run, builds are good and code coverage are as follows:

Tests:  1217 passing (5s)

## Coverage summary 
Statements   : 100% ( 4119/4119 )
Branches     : 89.63% ( 1755/1958 )
Functions    : 99.82% ( 1113/1115 )
Lines        : 100% ( 3934/3934 )

## Todos

- [x] Tests
- [x] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/common/src/server/components/ServerLoader.ts
+++ b/packages/common/src/server/components/ServerLoader.ts
@@ -398,6 +398,10 @@ export abstract class ServerLoader implements IServerLifecycle {
     await this.callHook("$beforeInit");
     await this.callHook("$onInit");
 
+    this.settings.forEach((value, key) => {
+      this.injector.logger.debug(`settings.${key} =>`, value);
+    });
+
     this.injector.logger.info("Build providers");
 
     await loadInjector(this.injector);
--- a/packages/common/src/server/components/ServerLoader.ts
+++ b/packages/common/src/server/components/ServerLoader.ts
@@ -398,10 +398,6 @@ export abstract class ServerLoader implements IServerLifecycle {
     await this.callHook("$beforeInit");
     await this.callHook("$onInit");
 
-    this.settings.forEach((value, key) => {
-      this.injector.logger.info(`settings.${key} =>`, value);
-    });
-
     this.injector.logger.info("Build providers");
 
     await loadInjector(this.injector);


pirouet:
I encountered this personally developing an application against a PCI Compliant service therefore the need to NOT print these settings out is paramount

packages/common/src/server/components/ServerLoader.ts
```diff
@@ -398,10 +398,6 @@ export abstract class ServerLoader implements IServerLifecycle {
     await this.callHook("$beforeInit");
     await this.callHook("$onInit");
 
-    this.settings.forEach((value, key) => {
-      this.injector.logger.info(`settings.${key} =>`, value);
```
Romakita: intead of remove log we can simply change the level by using debug ?
```typescript
this.injector.logger.debug 
```

Romakita:
Otherwise, I agree with you, all configuration shouldn't displayed in production mode (only debug)
Thanks for you PR :)

packages/common/src/server/components/ServerLoader.ts
pirouet: I can't approve; are you able to accept the requested change @Romakita ?

Romakita: why ?  Debug mode doesn't print any information by default. 

But ok, I propose to change the code by: 
```typescript
if (this.settings.logger.disableRoutesSummary) { // disable also route (or we can use another keyword)
   const {logger} =  this.injector
   
   logger.info('RootDir:', this.settings.rootDir)
   logger.info('Mount:', this.settings.mount)
   logger.info('ComponentScan:', this.settings.componentScan)
}
```
Theses informations are really important to debug application when a problem occurs about path resolution (initial settings are not enough to solve this problem). Other keys are not important for us.

Is it good for you ?

pirouet: Sorry, I meant GitHub wouldn't let me hit the "Approve" checkbox, I was happy with your suggestion of changing it from `.info` to `.debug`.

I'll amend the code and push just now

pirouet:
I have committed a modified version to make it `.debug` as per your suggestion

PR #634, opened 2019-08-29
Added flag to indicate not to switch file extensions

### LINKED ISSUE #477: Bootstraping problems when working with Jest/ts-jest
## Informations

Type | Version
---|---
Bug| 4.x
****

## Description
Hello,

When I bootstrap the server while using jest\ts-jest I've noticed that none of my controllers are mounted. 

I debugged into the loader and I saw that [next code](https://github.com/Romakita/ts-express-decorators/blob/05a67f3f90bbdb2d33edc4fd821101f5a5e738b6/packages/common/src/server/components/ServerLoader.ts#L610) checks for 'require.extensions[".ts"]'. But because ts-jest uses other way to compile typescript files into javascript this entry does not exit. The result of it is that pattern is changed from '...*.ts' into '...*.js'. But because those files do not exist nothing is mounted. 

When I changed (in debug) the patterns back to '...*.ts' everything worked fine.

```
  static cleanGlobPatterns(files: string | string[], excludes: string[]): string[] {
    excludes = excludes.map((s: string) => "!" + s.replace(/!/gi, ""));

    return []
      .concat(files as any)
      .map((file: string) => {
        if (!require.extensions[".ts"]) {
          file = file.replace(/\.ts$/i, ".js");
        }

        return Path.resolve(file);
      })
      .concat(excludes as any);

```

The only way it did work for me is only when I run it on compiled files.
 
Thank you!

TomerAmirV:
## Informations

Type | Breaking change
---|---
fixes #477 

****

## Description
When using `jest` and `ts-jest`, the `require.extensions['.ts']` compiler is not set.
Easy solution, for now, could be to pass an environment variable that will prevent the file extension change.

## Todos

- [x] Tests
- [ ] Coverage
- [x] Documentation

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -87,7 +87,8 @@ module.exports = {
               {link: "/tutorials/templating.html", text: "Templating"},
               {link: "/tutorials/throw-http-exceptions.html", text: "Throw HTTP exceptions"},
               {link: "/tutorials/not-found-page.html", text: "Customize 404"},
-              {link: "/tutorials/aws.html", text: "AWS project"}
+              {link: "/tutorials/aws.html", text: "AWS project"},
+              {link: "/tutorials/jest.html", text: "Jest"}
             ]
           },
           {text: "4.x", link: "http://v4.tsed.io"}
@@ -156,7 +157,8 @@ module.exports = {
               "templating",
               "throw-http-exceptions",
               "not-found-page",
-              "aws"
+              "aws",
+              "jest"
             ]
           }]
         },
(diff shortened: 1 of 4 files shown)


TomerAmirV:
@Romakita What do you think about this fix?
I know it's primitive, but I think that it's the simplest way...

Romakita:
If it works why not :)

Romakita:
But TS_TEST must be described somewhere. Otherwise, nobody will use this env variable.

TomerAmirV:
Sure! where can I update the docs?

TomerAmirV:
I think this is how you add docs, if now, please tell me and I will fix 😄

Romakita:
perfect :)

TomerAmirV:
Thanks!
When can we expect this to be released? 😄

Romakita:
Yes sorry, I'd just forgot to merge your PR :)

PR #679, opened 2019-10-01
fix(seq): Remove bold wraps

9KwHCUDP:
## Informations

Type | Breaking change
---|---
Fix | No

****

## Description
Remove bold wraps, as they are not shown correctly in seq.


Romakita:
Thanks @9KwHCUDP :)

PR #690, opened 2019-10-25
chore: refuse to include source files into package

ArtemDerevnjuk:
## Informations

Type | Breaking change
---|---
Chore | No

****

## Description
There's no point in including source files into `npm` package. It not only affects on a package size, but also can lead to re-compile `.ts` files in `node_modules`.

### DIFF
--- a/tasks/gulp/repo.js
+++ b/tasks/gulp/repo.js
@@ -138,7 +138,8 @@ module.exports = {
       .src([
         `${packagesDir}/**`,
         `${packagesDir}/**/.npmignore`,
-        `!${packagesDir}/**/src/**/*.{js,js.map,d.ts}`,
+        `!${packagesDir}/**/src/**`,
+        `!${packagesDir}/**/test/**`,
         `!${packagesDir}/**/package-lock.json`,
         `!${packagesDir}/**/yarn.lock`,
         `!${packagesDir}/**/node_modules/**`


Romakita:
But if the source isn't included, the source mapping won't works correctly if any bug occurs in your code. 
But you have to exclude node_modules in tsconfig to avoid recompilation :) The problem is here ;)

```typescript
{
  "compilerOptions": {
    "target": "es2016",
    "lib": ["es2016"],
    "typeRoots": [
      "./node_modules/@types"
    ],
    "module": "commonjs",
    "moduleResolution": "node",
    "experimentalDecorators":true,
    "emitDecoratorMetadata": true,
    "allowSyntheticDefaultImports": true
  },
  "exclude": [
    "node_modules"  // don't forget this line
  ]
}
```

ArtemDerevnjuk:
Hi @Romakita ,

I have some remarks)

> The problem is here ;)

Just see following example: https://github.com/ArtemDerevnjuk/tsed-example. How you can see it doesn't solve that issue

> you have to exclude node_modules in tsconfig to avoid recompilation

`exclude` has direct influence on search source files. It doesn't effect on module resolution. It means, if the compiler needs some `.ts` or `.d.ts` files to compile your sources, it will include them regardless `exclude` directive.

https://github.com/Microsoft/TypeScript/wiki/FAQ#why-is-a-file-in-the-exclude-list-still-picked-up-by-the-compiler
https://www.typescriptlang.org/docs/handbook/module-resolution.html#how-typescript-resolves-modules

> If the source isn't included, the source mapping won't works correctly

It doesn't matter. First, `tsc` allows to include source code into your source maps, you just need to enable `inlineSources`. Second, the published packages don't have the config files and dev dependencies to make a useful debugging experience. Therforce, there is no way for users to modify and compile `.ts` files in `node_modules`.

Romakita:
Ok ;). I'll take a time to test your PR on my machine before to be sure all is fine. I'll check tomorrow.

See you :)

Romakita:
@ArtemDerevnjuk Thanks for your tips. It works like a charm ;)

Issue #603, opened 2019-07-22
Nothing is loaded in v5.23.0 or greater


ghmendonca:
## Information

- **Version:** ^5.23.0
- **Type:** Issue

## Description
I just updated the version to 5.24.0 and suddenly all my routes stopped working. Testing this more, I noticed that services aren't working either, since graphql is not working. I tried 5.21.1 and 5.22.0 and both works, so I'm assuming the bug was introduced in 5.23.0.

One more thing that I noticed is that the 5.23.0 and 5.24.0 versions works on Ubuntu but not on Windows, so this leads me to believe this commit #b82a924 updated a dependency that introduced this bug, but I'm just guessing

striquer:
I'm having the same problem on my local windows, but on AWS is working fine, nothing on the mount or componentScan properties are being loaded

Romakita:
Hum I think the problem come from globby dependencies. I'll try to find a fix ASAP.

Romakita:
Fast-glob change the behavior with forward-slash. It's a big breaking change on Windows. 
https://github.com/mrmlnc/fast-glob/releases/tag/3.0.0

ghmendonca:
Can you provide us a work around for this? I read the breaking changes and I couldn't see a way to work around ☹

ghmendonca:
I tried the following thing: `path.normalize(path.join(__dirname, '/services/**/*.ts')).replace(/\\/g, '/')` but it didn't work.. I can see that using this, it displays the right path, using only forward slashes and the right path, but still the server doesn't load anything

ghmendonca:
I left the exact same code as I tried in the comment before and rolled back to 5.22.0 and it works, so maybe this is something else?

Romakita:
I published a new version v5.24.1 with the fix :) (rollback to globby v9 )

ghmendonca:
Awesome! Thanks man 😁

PR #708, opened 2019-11-21
Fix Scanning on Windows Paths

sampaioletti:
Type | Breaking change
---|---
Fix| No

****

## Description
Regression of #603 due to globby dependency, causing build errors on windows as paths resolve with "\" rather than "/"

Currently failing one test need to know if this is the correct fix?

### DIFF
--- a/packages/common/src/server/utils/cleanGlobPatterns.ts
+++ b/packages/common/src/server/utils/cleanGlobPatterns.ts
@@ -10,7 +10,7 @@ export function cleanGlobPatterns(files: string | string[], excludes: string[]):
         file = file.replace(/\.ts$/i, ".js");
       }
 
-      return Path.resolve(file);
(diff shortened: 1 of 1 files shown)


sampaioletti:
Dependency was changed from 9 to 10 here

https://github.com/TypedProject/ts-express-decorators/blame/production/package.json#L78

Romakita:
It works :)

Romakita:
Can we merge it ?

sampaioletti:
Ok this is passing all tests on windows now. Sorry for not following conventional commits before, i'll do that in the future. (:

Romakita:
@sampaioletti Thanks for your PR ;)

sampaioletti:
No problem @Romakita thanks for the great project, least I could do.

PR #721, opened 2019-12-11
Added missing types and alter wording.

ThomasHambach:
Added missing types and alter wording.

## Informations

Type | Breaking change
---|---
Doc | No

****

## Description
* To use passport.js and it's types, we need to install `@types/passport` as a development dependency. There were also some types missing in the examples. 
* Request scope was pointing to singleton scope example code.

## Todos
- [x] Documentation


Romakita:
thanks a lot for your PR @ThomasHambach ;)

PR #730, opened 2020-01-02
Fix dotenv example snippet

kpapadatos:
I believe this snippet was copy-pasted from the npm-config example and line 9 was not updated.


Romakita:
thanks @kpapadatos ;)

PR #759, opened 2020-02-10
Fix getter of debug setting and type annotations of port settings

thorlacius:
Type | Breaking change
---|---
Fix|Yes
****

## Description
Since `httpPort` and `httpsPort` can be `false` (to disable the respective protocol), I propose adding `false` to the untagged union type of those properties (and the `httpPort` synonym port). The setter and getter of settings.debug disagreed, so I propose changing the getter to return true iff the log level is `"debug"`.

## Retrospective
Automate type checking to improve the quality of the type annotations.

### DIFF
--- a/packages/common/src/config/services/ServerSettingsService.ts
+++ b/packages/common/src/config/services/ServerSettingsService.ts
@@ -162,7 +162,7 @@ export class ServerSettingsService extends DIConfiguration {
   }
 
   get debug(): boolean {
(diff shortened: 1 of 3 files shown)


Romakita:
Thanks for your PR.

I reverted the change on the tsconfig. The strict mode introduce a lot of compilation error over all packages. I waste  few minutes to understand where come from the problem XD

See you

PR #790, opened 2020-03-31
[BUGFIX/GRAPHQL/URGENT] dataSources needs to be instantiated each apollo request

bsparks:
the createDatasources is executed once in the server config, the getDatasources needs to be provider scope instance per request. It needs to be inside the factory call or else the internal caching mechanism of RestDataSource will not be properly flushed and cause a memory leak.


Romakita:
I'll fix coverage for you :)

bsparks:
fantastic, thank you!

PR #802, opened 2020-04-13
feature(common): Converter service remove additional Model property w…

amourlanne:
…hen removeAdditionalProperty is enabled in server settings

## Informations

Type | Breaking change
---|---
Feature | Yes

****

## Description
Converter service remove additional Model properties when removeAdditionalProperty is enabled in server settings.

## Usage example

```ts
@ServerSettings({
    validationModelStrict: false,
    removeAdditionalProperty: true
})
export class Server extends ServerLoader {

}
```

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation

### DIFF
--- a/packages/common/src/converters/services/ConverterService.ts
+++ b/packages/common/src/converters/services/ConverterService.ts
@@ -13,9 +13,11 @@ import {IConverter, IConverterOptions, IDeserializer, ISerializer} from "../inte
 @Injectable()
 export class ConverterService {
   private validationModelStrict = true;
+  private removeAdditionalProperty = false;
 
   constructor(private injectorService: InjectorService, @Configuration() configuration: Configuration) {
     this.validationModelStrict = configuration.get<boolean>("validationModelStrict");
+    this.removeAdditionalProperty = configuration.get<boolean>("removeAdditionalProperty");
   }
 
   /**
@@ -228,6 +230,10 @@ export class ConverterService {
   private convertProperty(obj: any, instance: any, propertyName: string, propertyMetadata?: PropertyMetadata, options?: any) {
     this.checkStrictModelValidation(instance, propertyName, propertyMetadata);
 
+    if (this.hasRemoveAdditionalPropertyEnabled() && propertyMetadata === undefined) {
+      return;
+    }
+
     propertyMetadata = propertyMetadata || ({} as any);
 
     let propertyValue = obj[propertyMetadata!.name] || obj[propertyName];
@@ -306,5 +312,13 @@ export class ConverterService {
     }
 
     return false;
+  }
+
+  /**
+   *
+   * @returns {boolean}
+   */
+  private hasRemoveAdditionalPropertyEnabled(): boolean {
+    return this.removeAdditionalProperty;
   }
 }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -98,6 +98,7 @@ With [dotenv](https://www.npmjs.com/package/dotenv):
 * `swagger` &lt;Object&gt;: Object configure swagger. See more on [Swagger](/tutorials/swagger.md).
 * `routers` &lt;object&gt;: Global configuration for the Express.Router. See express [documentation](http://expressjs.com/en/api.html#express.router).
 * `validationModelStrict` &lt;boolean&gt;: Use a strict validation when a model is used by the converter. When a property is unknown, it throw a `BadRequest` (see [Converters](/docs/converters.md)). By default true.
+* `removeAdditionalProperty` &lt;boolean&gt;: When strict validation is disabled. Remove additional property when a model is used by the converter. By default false.
 * `logger` &lt;[ILoggerSettings](/api/common/config/interfaces/ILoggerSettings.md)&gt;: Logger configuration.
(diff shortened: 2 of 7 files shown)


amourlanne:
Linked to [issue 470](https://github.com/TypedProject/tsed/issues/470).

Romakita:
Hello @amourlanne 
Thanks for your PR. 
I do not question the need for your PR which is really good :) But I would like us to go further ^^.
I would like to remove the validationModelStrict in futur to let AJV/JOI/ClassValidator to check this part. But now, I have two props related to ConvertersService, and this two props have the same goal. We can't use additionalProperty if the validationModelStrict is true.

Maybe refactoring the code to have only one option could be better :)

Example:
```
@ServerSettings({
  validationModelStrict: true // deprecated => converter.additionalProperty = true
  converter: { // or jsonmapper?
     additionalProperty: 'error' || 'accept' || 'ignore' 
     // futur option can be added to this props
  }
})
```

In second time we can add `@AdditionalProperty(boolean)` decorator. That allow developer to change the converter service behavior :).

What do you think ?
Romain

packages/common/src/converters/services/ConverterService.ts
```diff
@@ -13,9 +13,11 @@ import {IConverter, IConverterOptions, IDeserializer, ISerializer} from "../inte
 @Injectable()
 export class ConverterService {
   private validationModelStrict = true;
+  private removeAdditionalProperty = false;
 
   constructor(private injectorService: InjectorService, @Configuration() configuration: Configuration) {
     this.validationModelStrict = configuration.get<boolean>("validationModelStrict");
+    this.removeAdditionalProperty = configuration.get<boolean>("removeAdditionalProperty");
```
Romakita, line 20: if you're agree with my global comment:

```typescript
this.converterOptions = configuration.get<boolean>("converter") || {
   additionalProperty: validationModelStrict ? "error" : "accept"
};


```

packages/common/src/converters/services/ConverterService.ts
```diff
@@ -228,6 +230,10 @@ export class ConverterService {
   private convertProperty(obj: any, instance: any, propertyName: string, propertyMetadata?: PropertyMetadata, options?: any) {
     this.checkStrictModelValidation(instance, propertyName, propertyMetadata);
 
+    if (this.hasRemoveAdditionalPropertyEnabled() && propertyMetadata === undefined) {
```
Romakita, line 233:
```typescript
// rename checkStrictModelValidation by skipAdditionalProperty
if (skipAdditionalProperty(instance, propertyName, propertyMetadata)){
  return
}
```

New implementation:
```typescript
private skipAdditionalProperty(instance: any, propertyKey: string | symbol, propertyMetadata: PropertyMetadata | undefined) {
   const model = classOf(instance)
   if (model !== Object || propertyMetadata === undefined) {
      return false // Current instance isn't a Model
   }  
    
   // this part can be extracted to another method like getAdditionalPropertyLevel ?
   let additionalPropertyLevel = this.convertOptions.additionalProperty

   if (model !== Object) {
      if (Store.from(target).has('modelStrict')) {
        additionalPropertyLevel = !!Store.from(target).get("modelStrict")  ? 'error' : 'accept';
      }
   }
   // end getAdditionalPropertyLevel?
    
   switch(additionalPropertyLevel) {
       case "error":
          throw new UnknownPropertyError(model, propertyKey);
       case "ignore":
          return true;
       default:
       case "accept":
          return false;
   }
}
```

amourlanne:
Hello, your solution is better.
I have to make choice for the weight of validationModelStrict option and converter.additionalProperty option.

I choose this
| field | type |override | values |
|-------|-------|----------|--------------------------------|
| **validationModelStrict** (default) |  option |          | true =>"error", false =>"accept" (default) |
| **converter.additionalProperty** | option | validationModelStrict  | "error", "accept", "ignore"|
|  **ModelStrict** | decorator |  validationModelStrict and converter.additionalProperty | true =>"error", false =>"accept" |
| **AdditionalProperty** | decorator | validationModelStrict, converter.additionalProperty and  ModelStrict |  "error", "accept", "ignore" |

Is that right for you ? otherwise, what choices would you make to manage the weight of both options (and decorators).

Romakita:
Hello @amourlanne,

The @AdditionalProperties must follow the JsonSchema spec. It means, his signature accept a boolean or a JsonSchema. Implement this decorator can be done in another PR because, I work on this decorator for the new package @tsed/schema.
And I want to remove the @ModelStrict in favor of the Json schema validation. ConvertersService in future will just serialize and deserialize the models.

So for this PR you can follow your spec:
| field | type |override | values |
|-------|-------|----------|--------------------------------|
| **validationModelStrict** (default) |  option |          | true =>"error", false =>"accept" (default) |
| **converter.additionalProperties** | option | validationModelStrict  | "error", "accept", "ignore"|
|  **ModelStrict** | decorator |  validationModelStrict and converter.additionalProperty | true =>"error", false =>"accept" |

additionalProperties will define globally if we have to emit error, or accept/ignore properties.
@amourlanne ModelStrict will be deprecated when @AdditionalProperties will be implemented :)

See you !

PR #928, opened 2020-09-20
Objection.js plugin (WIP)

stefanvanherwijnen:
I started working on a @tsed/objection plugin to integrate [Objection.js](https://vincit.github.io/objection.js/) into Ts.ED.
Currently it's just very basic functionality based on the gist (https://gist.github.com/Romakita/32da7db47264f29bb535c6546459de9d)

A few questions:
- What functionality do I need to add to integrate it into Ts.ED the correct way (think services, providers etc)?
- Knex depends on database connection packages (pg, sqlite, mysql etc) which are not installed by default. Because the database connection is created inside the plugin, installing them into the main project does not work. How should I provide support for all databases without actually installing all the required packages?
- Currently the `Model` class is re-exported inside the plugin to make sure the same `Model` is used in the project and the plugin. This seems to work but I am wondering if this is the best way?

I had a look at the TypeORM package but that one seems much more complicated as it offers multiple connections etc, which is not possible with knex by default. I'll finish the plugin, but some pointers and feedback would be appreciated.

----------------------------------------------------------------------
To create a connection:
```ts
import {Configuration, registerProvider} from "@tsed/di";
import {setupObjectionJs} from "@tsed/objection";
import Knex, {Config} from "knex";

export const CONNECTION = Symbol.for("DEFAULT_CONNECTION");
export type CONNECTION = ReturnType<Knex>;

registerProvider({
  provide: CONNECTION,
  deps: [Configuration],
  useFactory(configuration: Configuration) {
    const settings = configuration.get<Config[]>("knex")!;
    const connectionOptions = settings[0];

    return setupObjectionJs(connectionOptions!);
  },
});
```
`models/User.ts`:
```ts
import {Required, MinLength, MaxLength} from "@tsed/common";
import {Model, Entity} from "@tsed/objection";

@Entity("users")
export class User extends Model {
  @Required()
  @MinLength(1)
  @MaxLength(25)
  name: string;
}
```

### DIFF
--- a/docs/tutorials/objection.md
+++ b/docs/tutorials/objection.md
@@ -1,0 +1,70 @@
+---
+meta:
+ - name: description
+   content: Use Objection.js with Express, TypeScript and Ts.ED.
+ - name: keywords
+   content: ts.ed express typescript objection.js node.js javascript decorators
+---
+# Objecion.js
+Install the dependencies:
+`yarn add @tsed/objection objection knex`
+
+Add to your server:
+`import "@tsed/objection"`
+
+Add a `knex` configuration to your TSed configuration (see: http://knexjs.org/#Installation-client for options).
+E.g.:
+```typescript
+@Configuration({
+  ...
+  knex: {
+    client: 'sqlite3',
+    connection: ':memory:'
+  }
+})
+```
+Setup a connection provider:
+```typescript
+import {Configuration, registerProvider} from "@tsed/di";
+import {createConnection} from "@tsed/objection";
+import Knex, {Config} from "knex";
+
+export const CONNECTION = Symbol.for("DEFAULT_CONNECTION");
+export type CONNECTION = ReturnType<Knex>;
+
+registerProvider({
+  provide: CONNECTION,
+  deps: [Configuration],
+  async useAsyncFactory(configuration: Configuration) {
+    const connectionOptions = configuration.get<Config>("knex")!;
+
+    if (connectionOptions) {
+      try {
+        return createConnection(connectionOptions!);
+      } catch (er) {
+        console.error(er);
+      }
+    }
+
+    return {};
+  }
+});
+```
+
+## Usage
+You can use the `Entity` decorator to create your models and make them work with Objection.js. `Entity` expects the table name as its argument.
+
+```typescript
+import {Required, MinLength, MaxLength, Inject} from "@tsed/common";
+import {Entity} from "@tsed/objection";
+import {Model} from 'objection';
+
+@Entity("users")
+export class User extends Model {
+  id: number
+  @Required()
+  @MinLength(1)
+  @MaxLength(25)
+  name: string;
+}
+```
--- a/packages/objection/package.json
+++ b/packages/objection/package.json
@@ -8,16 +8,36 @@
     "build": "tsc --build tsconfig.compile.json"
   },
   "dependencies": {
-    "knex": "^0.21.5",
-    "objection": "2.2.3",
-    "pg": "^8.3.3",
-    "sqlite3": "^5.0.0",
     "tslib": "1.11.0"
   },
   "private": false,
   "devDependencies": {
-    "@tsed/common": "5.65.2",
-    "@tsed/core": "5.65.2"
+    "@tsed/ajv": "^5.65.3",
+    "@tsed/common": "^5.65.3",
+    "@tsed/core": "5.65.2",
+    "ajv": "^6.12.5",
+    "knex": "^0.21.5",
+    "mssql": "^6.2.2",
+    "mysql": "^2.18.1",
+    "mysql2": "^2.2.5",
+    "objection": "2.2.3",
+    "oracledb": "^5.0.0",
+    "pg": "^8.3.3",
+    "sqlite": "^4.0.14",
+    "sqlite3": "^5.0.0"
   },
-  "peerDependencies": {}
+  "peerDependencies": {
+    "@tsed/ajv": "^5.65.3",
+    "@tsed/common": "5.65.2",
+    "ajv": "^6.12.5",
+    "knex": "^0.21.5",
+    "mssql": "^6.2.2",
+    "mysql": "^2.18.1",
+    "mysql2": "^2.2.5",
+    "objection": "2.2.3",
+    "oracledb": "^5.0.0",
+    "pg": "^8.3.3",
+    "sqlite": "^4.0.14",
+    "sqlite3": "^5.0.0"
+  }
 }
--- a/packages/objection/package.json
+++ b/packages/objection/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "@tsed/objection",
+  "version": "5.65.2",
+  "description": "Objection.js package for Ts.ED framework",
+  "main": "./lib/index.js",
+  "typings": "./lib/index.d.ts",
+  "scripts": {
+    "build": "tsc --build tsconfig.compile.json"
+  },
+  "dependencies": {
+    "knex": "^0.21.5",
+    "objection": "2.2.3",
+    "pg": "^8.3.3",
+    "sqlite3": "^5.0.0",
+    "tslib": "1.11.0"
+  },
+  "private": false,
+  "devDependencies": {
+    "@tsed/common": "5.65.2",
+    "@tsed/core": "5.65.2"
+  },
+  "peerDependencies": {}
+}
--- a/packages/objection/src/decorators/Entity.ts
+++ b/packages/objection/src/decorators/Entity.ts
@@ -1,18 +1,21 @@
-import {nameOf} from "@tsed/core";
 import {getJsonSchema} from "@tsed/common";
 
-export function Entity (tableName: string): ClassDecorator {
-  return (target) => {
-    Object.defineProperty(target, 'tableName', {
-      get(){
-        return tableName || (nameOf(target) + 's')
-      }
-    })
-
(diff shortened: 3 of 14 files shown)


packages/objection/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "@tsed/objection",
+  "version": "5.65.2",
+  "description": "Objection.js package for Ts.ED framework",
+  "main": "./lib/index.js",
+  "typings": "./lib/index.d.ts",
+  "scripts": {
+    "build": "tsc --build tsconfig.compile.json"
+  },
+  "dependencies": {
+    "knex": "^0.21.5",
```
Romakita, line 11: In fact, you don't need to keep theses dependencies here (knex, objection, pg, sqlite3)
This is my error when a created typeorm package. Because typeorm release often a new package, isn't easy to maintain the package.json up-to-date. So I moved in v6, typeorm to devDependencies (and peerDependencies also) section and in the readme, I added the instruction to install correctly the tsed package. 
It'll solve/prevent many issues :) but we need to correctly description the installation guide ;)

example:

To install @tsed/objection you have to install the following packages:
```typescript
npm i --save objection knex @types/objections @types/knex @tsed/objection
npm i --save pg  // change by the expected database
```

packages/objection/src/decorators/Entity.ts
```diff
@@ -0,0 +1,18 @@
+import {nameOf} from "@tsed/core";
+import {getJsonSchema} from "@tsed/common";
+
+export function Entity (tableName: string): ClassDecorator {
+  return (target) => {
+    Object.defineProperty(target, 'tableName', {
+      get(){
+        return tableName || (nameOf(target) + 's')
```
Romakita, line 8: For: (nameOf(target) + 's') 
I propose to use a [pluralize](https://www.npmjs.com/package/pluralize) npm module (as dependencies) to prevent automatic naming error ^^. What do you think?

packages/objection/src/index.ts
```diff
@@ -0,0 +1,12 @@
+import "@tsed/common";
+import {Config} from "knex";
+
+declare global {
+  namespace TsED {
+    interface Configuration {
+      knex: Config[];
```
Romakita, line 7: Ok if Objection support only one database, we don't need to provide an array interface, it could be confusing for your users ;)

packages/objection/src/utils/connect.ts
```diff
@@ -0,0 +1,20 @@
...
+  Model.knex(connection)
+  return connection
+}
+
+function createConnection(connectionOptions: Config) {
+  try {
+    const connection = Knex({...connectionOptions })
+    return connection;
+  } catch (err) {
+    $log.error('Knex connection failed')
```
Romakita, line 18: I think, catch the error connection isn't wanted by. Developpers prefer to crash the server when the connection fail ;). 

So the code can be simplied:
```typescript
function createConnection(connectionOptions: Config){
  const connection = Knex((connectionOptions)
  Model.knex(connection)
  return connection
}
```

packages/objection/src/utils/connect.ts
```diff
@@ -0,0 +1,20 @@
+import Knex, {Config} from "knex";
+import {$log} from "@tsed/logger";
+import {Model as ObjectionJSModel} from 'objection'
+
+export class Model extends ObjectionJSModel {}
```
Romakita, line 5: Ok. This is third point :)
If you declare objection as devDependencies and peerDependencies, re-export Model from Objection won't make sense, because I'll manipulate always the same Model class :) (problem solved by the peerDependencies ^^ )

Romakita (approved):

Romakita:
Hello @stefanvanherwijnen 
This is a good first PR, I'm very impatient to see the next commits 👍 
I answered to your differents questions by reviewing your PR and adding notes.
I approvedchange the PR  instead of request by error (wrong click XD). But it's not really important at this step.

So the your different questions will be solved by moving the dependencies to devDependencies and peerDependencies. Keep the caret "^" before the module version. Developpers will be free to install version of Objection,Knex, etc... for you, it'll be easier to maintain the @tsed/objections package. 

Again good job ;)
Romain

packages/objection/src/decorators/Entity.ts
stefanvanherwijnen: If you ask me I would just use `return tableName`. I copied the pluralization from your gist, but my personal opinion is to let the user define the table name and be responsible for the pluralization. `pluralize` will probably handle most edge cases, but I don't like having to load an entire module just to pluralize `person` to `people`.
But that's just my opinion.

stefanvanherwijnen:
Thanks for the feedback. I noticed the peerDependencies field and only after submitting the thought struck me that I should probably have used that.
I'll fix the things you mentioned later this week, and I'll also write some tests to check if everything works as it should.

Let me know if there is anything specific I need to take into account, otherwise I'll just keep it as simple as possible (e.g. support only one connection).

packages/objection/src/decorators/Entity.ts
Romakita: You’ve right stay simple :)

stefanvanherwijnen:
Could you help me out on how to retrieve the provided Ajv instance from `@tsed/ajv`?
I want to add
```
Object.defineProperty(target, "createValidator", {
  get() {
    return ajv;
  }
});
```
to Entity.ts, with `ajv` as an argument. However, I can't seem to figure out how to retrieve the ajv instance which is created by `registerProvider` in `@tsed/ajv`.

This is to support custom validators in the json schema.

Romakita:
Hello @stefanvanherwijnen 
The problem is the Ajv Instance built by Ts.ED is only available when you start the server with the container. Here, the createValidator method is static and live outside the DI container. You are only able to give your own Ajv instance.

See you :)

stefanvanherwijnen:
Makes sense. I'll first do some testing with custom validators and then try to see if I can incorporate them into an ajv instance globally. Just trying to prevent having multiple (equal) validators across the application.

docs/tutorials/objection.md
```diff
@@ -0,0 +1,70 @@
+---
+meta:
+ - name: description
+   content: Use Objection.js with Express, TypeScript and Ts.ED.
+ - name: keywords
+   content: ts.ed express typescript objection.js node.js javascript decorators
+---
+# Objecion.js
```
Romakita, line 8: Objection :)

docs/tutorials/objection.md
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+
+    return {};
+  }
+});
+```
+
+## Usage
+You can use the `Entity` decorator to create your models and make them work with Objection.js. `Entity` expects the table name as its argument.
```
Romakita, line 55: You can replace `Entity` by @@Entity@@ to link automatically the Entity to his Api reference page :)

Example:
Capture d’écran 2020-10-29 à 09 24 56

docs/tutorials/objection.md
```diff
@@ -0,0 +1,70 @@
...
+
+@Entity("users")
+export class User extends Model {
+  id: number
+  @Required()
+  @MinLength(1)
+  @MaxLength(25)
+  name: string;
+}
+```
```
Romakita, line 70: Can you add theses lines please:


```
## Author 

<GithubContributors :users="['stefanvanherwijnen']"/>

## Maintainers

<GithubContributors :users="['stefanvanherwijnen']"/>

<div class="container--centered container--padded">
<a href="/contributing.html" class="nav-link button">
 Become maintainer
</a>
</div>
```
 
This code will display you github account ;)

docs/tutorials/objection.md
```diff
@@ -0,0 +1,70 @@
...
+   content: ts.ed express typescript objection.js node.js javascript decorators
+---
+# Objecion.js
+Install the dependencies:
+`yarn add @tsed/objection objection knex`
+
+Add to your server:
+`import "@tsed/objection"`
+
+Add a `knex` configuration to your TSed configuration (see: http://knexjs.org/#Installation-client for options).
```
Romakita, line 15: Ts.ED

packages/objection/src/decorators/IdColumn.ts
```diff
@@ -0,0 +1,7 @@
+// export function IdColumn (target: Object, propertyKey: string): PropertyDecorator {
```
Romakita, line 1: We can remove this file?

Romakita (changes requested):

Romakita:
Hello @stefanvanherwijnen 
Do you need help to finish this plugin ?
I also added some request changes :) 

Another point, do you need this feature on v5 branch? 
See you
Romain

stefanvanherwijnen:
My plan was to finish it by using it in a project and see what's still missing, but I haven't had the time unfortunately.
I'll fix the things you mentioned, and finish the rest, but I can't tell you when it will be done.

You can skip this for v5. I just want to make sure it works properly before I finish the PR.

Romakita:
I suggest to merge this PR and add a text on the readme and documentation to tell that this module is under work in progress / alpha. It allow you to test the module :)

stefanvanherwijnen:
Yes, good point. I was trying to integrate it into the CLI but without an upstream package you quickly get stuck. The PR can be merged, but as you said it is still in alpha. I'll make a new PR for any additions.

Romakita:
Yes no worries about that, I'll configure the package under alpha :) and add the badge on documentation. 
I'll trigger you when the package is available ;)

Romakita (approved):

Issue #1057, opened 2020-11-07
[V6] [BUG] @Ignore doesn't seem to be passed down through sub classes


from-nibly:
## Information

- **Version:** 6.3.1

When adding the @Ingore annotation to a super class serializing a Sub class doesn't seem to ignore the property and it is still serialized into the response.

## Example

```ts
import {  Ignore } from '@tsed/schema';
class A {
  @Ignore()
  ignoreMe: string
}

class B extends A {
  foo: string
}

...
@Returns(200, B)
async getThing() {
   return plainObjectToClass({ foo: 'bar', ignoreMe: 'testing'})
}
```

Response
```json
{ "foo": "bar", "ignoreMe": "testing"}
```

## Acceptance Criteria
ignoreMe should be removed from the json response

Romakita:
Hello @from-nibly 
The PR is ready :) Thanks for your issue!
Romain

PR #1060, opened 2020-11-08
fix(#1057): Value returned by an endpoint without @Returns is now correctly serialized

Romakita:
Closes: #1057


Romakita:
/ rebase

PR #1068, opened 2020-11-11
feat(view): add load consolidate requires options

adepanges:
to handle adding requires in consolidation, i am using nunjucks for view engine and have problem when extends layout or include some view will catch by error, this changes to solve it

## Usage example

```typescript
import { Configuration } from "@tsed/common";
import nunjucks from "nunjucks";

@Configuration({
    views: {
        root: `${rootDir}/views`,
        viewEngine: "nunjucks",
        extensions: {
            "njk": "nunjucks"
        },
        options: {
            nunjucks: {
                requires: nunjucks.configure(`${rootDir}/views`)
            }
        }
    },
})
export default class ShopApp {}

```

reference https://github.com/tj/consolidate.js#template-engine-instances


Romakita:
Great PR. But I doesn't seen test to cover this usecase. 
You can add it in /packages/common/src/platform/services/PlatformViews.spec.ts please :)

Romakita (changes requested):
Missing unit test :)

adepanges:
hai @Romakita 

sorry I've never written a unit test

Should I write a new **_describe_**, or just add **_it(use case)_** for my changes?

and also how do I define a view file to test? I am looking for the **_views.ejs_** file but I can't find it
i want to make new **_view.njk_** also **_layouts/default.njk_** if it is ok

Romakita:
Hello @adepanges 

You just need to see if you configuration is correctly handled by the method.

Just add requires to initial config:

```typescript

  beforeEach(() =>
    PlatformTest.create({
      views: {
        extensions: {
          ejs: "ejs"
        },
        viewEngine: "ejs",
        options: {
          ejs: {
            global: "global",
            requires: "requires" /// add this
          }
        }
      }
    })
  );
```
Then add a it:

```typescript
 it("should load engine requires", async () => {
      const platformViews = PlatformTest.get<PlatformViews>(PlatformViews);

      expect(platformViews.consolidate.ejs.requires).to.equal('requires');
 });
```
Try it :)

Romakita:
@adepanges There are some issue on unit test, can you fix it please ?

adepanges:
yes checking it, try to run unit test in my local also

Romakita:
yes just run `yarn test` :)

packages/common/src/platform/services/PlatformViews.spec.ts
```diff
@@ -15,7 +15,8 @@ describe("PlatformViews", () => {
         viewEngine: "ejs",
         options: {
           ejs: {
-            global: "global"
+            global: "global",
+            requires: "requires"
           }
         }
```
Romakita, line 21: add just before the first afterEach this code to fix your unit test:

```typescript
  afterEach(() => {
    const platformViews = PlatformTest.get<PlatformViews>(PlatformViews);

    delete platformViews.consolidate.requires.ejs;
  });
```

Because we add fake requires, we need to clean the  `platformViews.consolidate.requires.ejs` after the unit test, to be sure the integration test works ^^.

Romakita (changes requested):

packages/common/src/platform/services/PlatformViews.spec.ts
adepanges: okey thanks let meed add it, and run test

Romakita (approved):

PR #1082, opened 2020-11-12
fix(schema): Added file upload support

dvarnai:
## Informations

Type | Breaking change
---|---
Fix | No

****

## Description

This commit tries to reintroduce file support. I'm not sure if this is the best way to do this, so please let me know if you'd like anything changed.

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/common/src/platform/decorators/multer/multipartFile.spec.ts
+++ b/packages/common/src/platform/decorators/multer/multipartFile.spec.ts
@@ -1,6 +1,6 @@
 import {MultipartFile, ParamMetadata, ParamTypes, PlatformMulterMiddleware, Post} from "@tsed/common";
 import {descriptorOf, Metadata, Store} from "@tsed/core";
-import {getSpec} from "@tsed/schema";
+import {getSpec, SpecTypes} from "@tsed/schema";
 import {expect} from "chai";
 import * as Sinon from "sinon";
 
@@ -12,29 +12,29 @@ describe("@MultipartFile()", () => {
   const sandbox = Sinon.createSandbox();
 
   describe("one file", () => {
-    it("should set endpoint metadata", () => {
-      // WHEN
-      class TestController {
-        @Post("/")
-        test(@MultipartFile("file1", 1) file: any) {}
-      }
+    // WHEN
+    class TestController {
+      @Post("/")
+      test(@MultipartFile("file1", 1) file: any) {}
+    }
 
-      // THEN
-      const store = Store.fromMethod(TestController, "test");
+    // THEN
+    const store = Store.fromMethod(TestController, "test");
 
-      expect(store.get(PlatformMulterMiddleware)).to.deep.eq({
-        fields: [
-          {
-            maxCount: 1,
-            name: "file1"
-          }
-        ]
-      });
+    expect(store.get(PlatformMulterMiddleware)).to.deep.eq({
+      fields: [
+        {
+          maxCount: 1,
+          name: "file1"
+        }
+      ]
+    });
 
-      const param = ParamMetadata.get(TestController, "test", 0);
-      expect(param.expression).to.eq("file1.0");
-      expect(param.paramType).to.eq(ParamTypes.FILES);
+    const param = ParamMetadata.get(TestController, "test", 0);
+    expect(param.expression).to.eq("file1.0");
+    expect(param.paramType).to.eq(ParamTypes.FILES);
 
+    it("should set endpoint metadata - OS2", () => {
       expect(getSpec(TestController)).to.deep.eq({
         paths: {
           "/": {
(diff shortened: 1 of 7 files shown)


dvarnai:
Currently, this seems to create the OpenApi specification properly, however, validation fails as AJV doesn't know format: binary, also it tries to validate PlatformMulterFile as a string and fails

packages/schema/src/domain/JsonOperation.ts
```diff
@@ -198,6 +198,16 @@ export class JsonOperation extends JsonMap<JsonOperationOptions> {
         if (parameter.get("in")) {
           if (parameter.get("in") === JsonParameterTypes.BODY) {
             bodyParameters.push(parameter);
+          } else if (parameter.get("in") === JsonParameterTypes.FILES) {
+            if (options.specType === SpecTypes.OPENAPI) {
+              const schema = parameter.$schema.get("type") === "array" ? parameter.$schema.itemSchema() : parameter.$schema;
```
Romakita, line 203: I think this code should be in JsonParameter class instead of Operation (in `JsonParameter.toJSON`).
Finally we just need to instrument correctly where the parameters will be pushed:
```typescript
if ([JsonParameterTypes.BODY, JsonParameterTypes.FILES].includes(parameter.get("in"))) {
    bodyParameters.push(parameter);
}
```

Romakita (changes requested):

Romakita:
Hello @dvarnai 

I suggest to create a unit test to cover you scenario:
> packages/schema/src/decorators/operations/in.spec.ts

```typescript
describe("@In", () => {
  it("JsonParameterTypes.FILES with OS2", () => {
     class MyController {
      @OperationPath("POST", "/")
      method(@In(JsonParameterTypes.FILES) @Name("file") file: string) {
      } 
    }

    const result = getSpec(Controller, {
      specType: SpecTypes.SWAGGER
    });

    expect(result).to.deep.equal({...});
  });

  it("JsonParameterTypes.FILES with OS3", () => {
     class MyController {
      @OperationPath("POST", "/")
      method(@In(JsonParameterTypes.FILES) @Name("file") file: string) {
      } 
    }

    const result = getSpec(Controller, {
      specType: SpecTypes.OPENAPI
    });

    expect(result).to.deep.equal({...});
  });
});
```

Romakita:
> Currently, this seems to create the OpenApi specification properly, however, validation fails as AJV doesn't know format: binary, also it tries to validate PlatformMulterFile as a string and fails. 

In your PR on don't see how the given schema to AJV can have a binary format. But I think if you move the code to JsonParameter, the problem will be solved.

In JsonParameter:
```typescript
   if (this.get("in") === JsonParameterTypes.FILES) {
      const isOpenApi = options.specType === SpecTypes.OPENAPI;

      const schema = {
        type: isOpenApi ? "string" : "file",
        format: isOpenApi ? "binary" : undefined
      };

      if (jsonSchema.type === "array") {
        jsonSchema.items = cleanObject({
          ...jsonSchema.items,
          ...schema
        });

        parameter.schema = jsonSchema;
      } else {
        parameter.schema = cleanObject({
          ...jsonSchema,
          ...schema
        });
      }
    }
```

> Furthermore, ".0" gets appended to the name of the field in swagger.json. Sorry I couldn't be more useful

Hum... I haven't solution yet. I propose to fix that later :). Wait and see the requested change and after we will find a solution.

packages/schema/src/domain/JsonOperation.ts
Romakita: In JsonParameter, line just after `parameter.required = parameter.required || this.get("in") === "path";` line-60:

```typescript
if (this.get("in") === JsonParameterTypes.FILES) {
      const isOpenApi = options.specType === SpecTypes.OPENAPI;

      const schema = {
        type: isOpenApi ? "string" : "file",
        format: isOpenApi ? "binary" : undefined
      };

      if (jsonSchema.type === "array") {
        jsonSchema.items = cleanObject({
          ...jsonSchema.items,
          ...schema
        });

        parameter.schema = jsonSchema;
      } else {
        parameter.schema = cleanObject({
          ...jsonSchema,
          ...schema
        });
      }
    }
```

Romakita: I think by this way, the jsonSchema generated for AJV won't fail, because, format will be generated only for OS2 and OS3 :)

dvarnai:
Thanks for the review. I just pushed another commit. It seems we can't do this purely in JsonParameter.ts as the $schema is added directly as a property to JsonOperation, so now I'm manually calling JsonParameter.toJSON() in JsonOperation when building the schema.

Romakita:
To solve issue with the expression we need to update the unit test in the first time:

> See: packages/common/src/platform/decorators/multer/multipartFile.spec.ts

```diff
      expect(getSpec(TestController)).to.deep.eq({
        paths: {
          "/": {
            post: {
              consumes: ["multipart/form-data"],
              operationId: "testControllerTest",
              parameters: [
+                {
+                  in: "body",
+                  name: "body",
+                  required: false,
+                  schema: {
+                    properties: {
+                      "file1": {   // name without index
+                        $ref: "file"
+                      }
+                    },
+                    type: "object"
+                  }
+                }
              ],
              responses: {
                "400": {
                  description:
                    "<File too long | Too many parts | Too many files | Field name too long | Field value too long | Too many fields | Unexpected field>  [fieldName] Example: File too long file1",
                  schema: {
                    type: "string"
                  }
                }
              },
              tags: ["TestController"]
            }
          }
        },
        tags: [
          {
            name: "TestController"
          }
        ]
      });
    });
```

Then we need to fix JsonOperation to build the correct spec. To solve it, we can add `JsonParameter.getName` method, which will return the correct name.

```typescript
class JsonParameter {
  getName() {
    const name = this.get("name");

    if (this.get("in") === JsonParameterTypes.FILES) {
      return name.split(".")[0];
    }

    return name;
  }
}
```
And finally we can use this new method in JsonOperation file (fn: buildSchemaFromBodyParameters):

```diff
  parameters.forEach((parameter) => {
-    const name = parameter.get('name')
+    const name = parameter.getName()
```

What do you think :)?

dvarnai:
PR ready :)

packages/common/src/platform/decorators/multer/multipartFile.spec.ts
```diff
@@ -12,36 +12,169 @@ describe("@MultipartFile()", () => {
...
+    }
 
-      // THEN
-      const store = Store.fromMethod(TestController, "test");
+    // THEN
+    const store = Store.fromMethod(TestController, "test");
 
-      expect(store.get(PlatformMulterMiddleware)).to.deep.eq({
-        fields: [
+    expect(store.get(PlatformMulterMiddleware)).to.deep.eq({
```
Romakita, line 24: expect must be alway in a it :) not directly in a describe.

```typescript
it("should store metadata", () => {
     const store = Store.fromMethod(TestController, "test");
     expect(store.get(PlatformMulterMiddleware)).to.deep.eq({
      fields: [
        {
          maxCount: 1,
          name: "file1"
        }
      ]
    });
});

it("should create param with the right configuration", () => {
    const param = ParamMetadata.get(TestController, "test", 0);
    expect(param.expression).to.eq("file1.0");
    expect(param.paramType).to.eq(ParamTypes.FILES);
});

```

packages/common/src/platform/decorators/multer/multipartFile.spec.ts
```diff
@@ -12,36 +12,169 @@ describe("@MultipartFile()", () => {
...
+
+  describe("multiple file schema", () => {
+    // WHEN
+    class TestController {
+      @Post("/")
+      test(@MultipartFile("file1", 4) file: any[]) {}
+    }
 
+    // THEN
+    const store = Store.fromMethod(TestController, "test");
```
Romakita, line 138: same thing :)

Romakita (changes requested):

Romakita:
Thanks @dvarnai :)
Just to small change and it'll be ok :)

packages/common/src/platform/decorators/multer/multipartFile.spec.ts
dvarnai: ah you're right, sorry about that

Romakita: No worries ;)

Romakita (approved):

PR #1283, opened 2021-03-14
fix(schema): Avoid forwarding groups to children Model

Romakita:
When we using groups and Ref decorators, an issue can occurs because groups options generate wrong children models.

Now the groups are only applied on the current model (parent model only).


julianmills:
Hi @Romakita, I'm trying to work out why I can't get groups in nested objects to work in my documentation, and wondering if this change is preventing the expected behaviour: I have models that have properties such as `id` decorated with `@Groups("!creation")` that have children with the same. It looks like this change means that on the creation group the id property of the parent is correctly hidden, but not on the child, even though that is also being created. Is there a way of excluding child properties for a group, or is that now impossible?

PR #1337, opened 2021-04-07
feat(before-deserialize-decorator): Add BeforeDeserialize decorator

### LINKED ISSUE #1336: [Feat] BeforeDeserialize decorator
## Informations

BeforeDeserialize will allow to implement custom handler which will be called before the JsonMapper. It will be possible to 
change some values in the data and also implement custom validator.

## Example
```typescript
@BeforeDeserialize((data: Record<string, unknown>) => {
   if (data.prop === 'VALUE' && data.prop2 !== 'VALUE2') {
     throw BadRequest('MyMessage'); /// or throw a ValidationError
   }

   return data
})
class MyClass {}
```

## Solution

The new decorator will be added to the `@tsed/json-mapper` package and will be similar to the `@OnDeserialize` decorator. The `@Ignore` use the `hooks` API to register handler.

See decorator: https://github.com/TypedProject/tsed/blob/production/packages/json-mapper/src/decorators/onDeserialize.ts

When the decorator is implemented, we need to create the `alterBeforeDeserialize` function to call the `beforeDeserialize` hook. This function will provide to other packages the right way to call the hook. See example: https://github.com/TypedProject/tsed/blob/production/packages/schema/src/hooks/alterIgnore.ts

expected signature
```
alterBeforeDeserialize(data, store.schema, options);
```

> Create the hooks directory in the @tsed/json-mapper package.

The last step is to use the new hook in the right place:
https://github.com/TypedProject/tsed/blob/production/packages/json-mapper/src/utils/deserialize.ts#L87

```diff
- const out: any = new type(src);
+ const out: any = new type(alterBeforeDeserialize(data, store.schema, options));
```

Finally, add integration test here: https://github.com/TypedProject/tsed/tree/production/packages/json-mapper/test/integration

## Acceptance criteria

- [x] BeforeDeserialize and alterBeforeDeserialize decorator are exported
- [x] BeforeDeserialize is only applicable on class
- [x] The hook can alter the given object properties 
- [x] The hook can throw an error
- [x] Unit/Integration tests cover the code

xCryzed:
## Information

Type | Breaking change
---|---
Feature | No

****

## Description
Closes #1336.

### DIFF
--- a/packages/json-mapper/src/utils/deserialize.ts
+++ b/packages/json-mapper/src/utils/deserialize.ts
@@ -3,6 +3,7 @@ import {alterIgnore, getProperties, JsonEntityStore, JsonHookContext, JsonSchema
 import "../components";
 import {JsonMapperContext} from "../domain/JsonMapperContext";
 import {getJsonMapperTypes} from "../domain/JsonMapperTypesContainer";
+import {alterBeforeDeserialize} from "../hooks/alterBeforeDeserialize";
 import {JsonMapperMethods} from "../interfaces/JsonMapperMethods";
 
 export interface JsonDeserializerOptions<T = any, C = any> extends MetadataTypes<T, C> {
@@ -84,7 +85,7 @@ export function plainObjectToClass<T = any>(src: any, options: JsonDeserializerO
 
   let keys = objectKeys(src);
   const additionalProperties = propertiesMap.size ? !!store.schema.get("additionalProperties") || options.additionalProperties : true;
-  const out: any = new type(src);
+  const out: any = new type(alterBeforeDeserialize(src, store.schema, options));
 
   propertiesMap.forEach((propStore) => {
     const key = options.useAlias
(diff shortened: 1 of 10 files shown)


packages/json-mapper/src/utils/deserialize.ts
```diff
@@ -84,7 +85,7 @@ export function plainObjectToClass<T = any>(src: any, options: JsonDeserializerO
 
   let keys = objectKeys(src);
   const additionalProperties = propertiesMap.size ? !!store.schema.get("additionalProperties") || options.additionalProperties : true;
-  const out: any = new type(src);
+  const out: any = new type(alterBeforeDeserialize(src, store.schema, options));
```
Romakita, line 88: Hum ok, maybe to avoid confusing behavior (and answer your question about the `return data`), the following code will more appropriate and throw an error if data isn't correctly returned:
```diff
-   const out: any = new type(alterBeforeDeserialize(src, store.schema, options));
+  src = alterBeforeDeserialize(src, store.schema, options);
+  const out: any = new type(src);
```
In this case if data isn't returned, the mapping will fail!

Romakita (changes requested):

packages/json-mapper/src/utils/deserialize.ts
xCryzed: done :)

Romakita (approved):

Romakita:
Great Job @xCryzed ;)

Romakita (approved):

PR #1339, opened 2021-04-08
fix: use seconds instead of expires for cache-control max-age value

jkrnak:
## Information

Type | Breaking change
---|---
Fix | No

****

## Description

The `max-age` value of the `cache-control` header should be a relative value to the request's time not an absolute timestamp.
Ie. to cache something for 30 seconds should have `cache-control: max-age=30`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#expiration
https://www.keycdn.com/blog/http-cache-headers#:~:text=downloaded%20every%20time.-,max%2Dage,for%20the%20next%2090%20seconds.

## Usage example

Nothing changes in the usage.

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation


Romakita (approved):

Romakita:
Thanks for your PR ;) @jkrnak

jkrnak:
Thanks for the quick merge @Romakita .
Amazing framework btw, I really love it! Keep up the great work!

PR #1359, opened 2021-04-18
docs: Add Command documentation

Romakita:


docs/docs/command.md
```diff
@@ -0,0 +1,305 @@
...
+export interface HelloCommandContext {
+  rawArgs: string[];
+}
+
+@Command({
+  name: "hello-command",
+  description: "Command description",
+  args: {},
+  options: {},
+  allowUnknownOption: false
```
superkaleider, line 220: in this example, should this be true?

Romakita: Yes :) fixed

PR #1382, opened 2021-05-05
feat: added ability to customize request handler

YaroslavAverianovAE:
## Information

Type | Breaking change
---|---
Feature | No

****

## Description
Added ability to hack `onRequest` handler.
Could be useful for context defining.

### DIFF
--- a/packages/common/src/platform/services/PlatformHandler.ts
+++ b/packages/common/src/platform/services/PlatformHandler.ts
@@ -12,6 +12,7 @@ import {
 } from "../../mvc";
 import {ValidationError} from "../../mvc/errors/ValidationError";
 import {HandlerContext} from "../domain/HandlerContext";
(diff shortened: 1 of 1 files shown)


packages/common/src/platform/services/PlatformHandler.ts
```diff
@@ -1,3 +1,4 @@
+import {PlatformContext} from "@tsed/common";
```
Romakita, line 1: The build will fail because you try to import a class from the same package. Change the path by a relative path.

Romakita (changes requested):

packages/common/src/platform/services/PlatformHandler.ts
YaroslavAverianovAE: Done

Romakita (approved):

PR #1490, opened 2021-08-16
OIDC

stefanvanherwijnen:
A few points:
- Because oidc-provider is a koa app with a catch-all route, mounting it at `/` intervenes with the TSed exception handler. This is fixed by mounting it at `/oidc` and rewriting the route for the discovery documentation.
- For Koa, oidc needs to be mounted differently which now works correctly (previously all requests would be handled by both the TSed and oidc-provider koa instance)
- There are no type declarations for koa-rewrite, which results in `test/app` not compiling. I couldn't figure out how to fix this.
- I cleaned the documentation and embedded the files from the `test/app` directory.
- I added documentation for a TLS proxy using Caddy.

I couldn't directly test the app/documentation as it does not compile, so if you know a fix that would be great. Because the code is now embedded in the docs, if the test app works the documentation is also correct.

The CLI plugin is almost finished, but I want to make sure everything is consistent along all documentation.
For example, the CLI initializes an app with default port 8083, and the documentation used 8081. I'd like to make sure that the CLI automates the steps in the documentation without modifications.

### DIFF
--- a/packages/oidc-provider/src/OidcModule.ts
+++ b/packages/oidc-provider/src/OidcModule.ts
@@ -1,5 +1,5 @@
 import {Inject, InjectorService, PlatformApplication} from "@tsed/common";
-import {Module} from "@tsed/di";
+import {Constant, Module} from "@tsed/di";
 import {OidcAdapters} from "./services/OidcAdapters";
 import {OidcJwks} from "./services/OidcJwks";
 import {OidcProvider} from "./services/OidcProvider";
@@ -10,6 +10,12 @@ import {OidcProvider} from "./services/OidcProvider";
 export class OidcModule {
   @Inject()
   app: PlatformApplication;
+
+  @Constant("PLATFORM_NAME")
+  platformName: string;
+
+  @Constant("oidc.path", "/")
+  basePath: string;
 
   @Inject()
   oidcProvider: OidcProvider;
@@ -23,11 +29,25 @@ export class OidcModule {
     }
   }
 
+  async $onRoutesInit() {
+    if (this.basePath !== "/") {
+      this.app.use(this.getRewriteMiddleware());
+    }
+  }
+
   async $afterRoutesInit() {
     if (this.oidcProvider.hasConfiguration()) {
       const provider = this.oidcProvider.get();
 
-      this.app.use(provider.callback());
+      switch (this.platformName) {
+        default:
+        case "express":
+          this.app.use(this.basePath, provider.callback());
+          break;
+        case "koa":
+          this.app.use(require("koa-mount")(this.basePath, provider.app));
+          break;
+      }
     }
   }
 
@@ -50,6 +70,16 @@ export class OidcModule {
         const host = injector.settings.getHttpPort();
         displayLog({protocol: "http", ...host});
       }
+    }
+  }
+
+  private getRewriteMiddleware() {
+    switch (this.platformName) {
+      default:
+      case "express":
+        return require("express-urlrewrite")("/.well-known/*", `${this.basePath}/.well-known/$1`);
+      case "koa":
+        return require("koa-rewrite")("/.well-known/(.*)", `${this.basePath}/.well-known/$1`);
     }
   }
 }
--- a/packages/oidc-provider/test/oidc.integration.spec.ts
+++ b/packages/oidc-provider/test/oidc.integration.spec.ts
@@ -8,6 +8,9 @@ import {rootDir} from "../../platform-express/test/app/Server";
 import {InteractionsCtrl} from "./app/controllers/oidc/InteractionsCtrl";
 import {Server} from "./app/Server";
 
+import {join} from "path";
+import {Accounts} from "./app/services/Accounts";
+
 const utils = PlatformTestUtils.create({
   rootDir,
   platform: PlatformExpress,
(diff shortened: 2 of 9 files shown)


Romakita:
Hello @stefanvanherwijnen
Thanks for the PR

Romakita:
Hello @stefanvanherwijnen 

Thanks for the PR.

> There are no type declarations for koa-rewrite, which results in test/app not compiling. I couldn't figure out how to fix this.   

The test app in packages/oidc ? Can you give me more details?

> Because the code is now embedded in the docs

I don't understand this point...

> The CLI plugin is almost finished, but I want to make sure everything is consistent along all documentation.
> For example, the CLI initializes an app with default port 8083, and the documentation used 8081. I'd like to make sure that the CLI automates the steps in the documentation without modifications.

Update documentation is possible to be aligned with the CLI ;)

stefanvanherwijnen:
> Hello @stefanvanherwijnen
> 
> Thanks for the PR.
> 
> > There are no type declarations for koa-rewrite, which results in test/app not compiling. I couldn't figure out how to fix this.
> 
> The test app in packages/oidc ? Can you give me more details?
> 

Yes. Running `yarn start` results in "src/OidcModule.ts:7:21 - error TS7016: Could not find a declaration file for module 'koa-rewrite'". In `src` I manually declared `koa-rewrite`, but it doesn't seem to get picked up when using `yarn start`. Ideally `yarn start` would spin up a working application on which the URLs defined in the documentation can be opened (e.g. the .well-known route).

> > Because the code is now embedded in the docs
> 
> I don't understand this point...

I mean using "<<< @/../packages/oidc-provider/src/domain/OidcSettings.ts" instead of copy-pasting the code into the documentation. This way it is easier to keep everything in sync without having to check everything manually.


I am working on fixing the integration test.

stefanvanherwijnen:
The coverage test seems to fail but I am not sure how to fix it.

Romakita:
> Yes. Running yarn start results in "src/OidcModule.ts:7:21 - error TS7016: Could not find a declaration file for module 'koa-rewrite'". In src I manually declared koa-rewrite, but it doesn't seem to get picked up when using yarn start. Ideally yarn start would spin up a working application on which the URLs defined in the documentation can be opened (e.g. the .well-known route).

You can simply use require instead of import from for this case ;)

> I mean using "<<< @/../packages/oidc-provider/src/domain/OidcSettings.ts" instead of copy-pasting the code into the documentation. This way it is easier to keep everything in sync without having to check everything manually.

Ok I see ;) I'm ok to use this way !!

stefanvanherwijnen:
Thanks, one more thing which does not seem to work: the `Post()` routes in the interactions do not seem to get registered when using koa. E.g. POST to `/interactions/123/login` results in a NOT_FOUND error with koa, but with express it results in SessionNotFound as expected. The main GET routes do work fine.

Any clue?

packages/oidc-provider/src/OidcModule.ts
```diff
@@ -23,11 +28,23 @@ export class OidcModule {
...
+    }
+  }
+
   async $afterRoutesInit() {
     if (this.oidcProvider.hasConfiguration()) {
       const provider = this.oidcProvider.get();
 
-      this.app.use(provider.callback());
+      if (this.platformName === "koa") {
+        this.app.use(mount("/oidc", provider.app));
```
Romakita: hum... let the user to configure the correct path from configuration could be a better options. Because today, /oidc isn't the default path and it will be a breaking change. no?

packages/oidc-provider/src/OidcModule.ts
```diff
@@ -1,16 +1,21 @@
...
 @Module({
   imports: [OidcProvider, OidcAdapters, OidcJwks]
 })
 export class OidcModule {
   @Inject()
   app: PlatformApplication;
 
+  @Constant("PLATFORM_NAME")
+  platformName: string;
+
```
Romakita, line 16:
```typescript
@Constant('oidc.path', '/')
private basePath: string;
```

packages/oidc-provider/src/OidcModule.ts
```diff
@@ -23,11 +28,23 @@ export class OidcModule {
     }
   }
 
+  async $onRoutesInit() {
+    if (this.platformName === "koa") {
```
Romakita:
```typescript
const urlRewrite = [this.basePath, ".well-known/$1"].join('/')

 if (this.platformName === "koa") {
    this.app.use(rewrite("/.well-known/(.*)", urlRewrite);
  } else {
    this.app.use(expressRewrite("/.well-known/*", urlRewrite));
  }
```

packages/oidc-provider/src/OidcModule.ts
Romakita:
```typescript
      if (this.platformName === "koa") {
        this.app.use(mount(this.basePath, provider.app));
      } else {
        this.app.use(this.basePath, provider.callback());
      }
```

packages/oidc-provider/src/OidcModule.ts
Romakita: Or maybe we can try to retrieve the path of the InteractionsCtrl ?

packages/oidc-provider/src/OidcModule.ts
Romakita:
```typescript
  $onReady() {
    if (this.oidcProvider.hasConfiguration()) {
      const {injector} = this;
      const {httpsPort, httpPort} = injector.settings;

      const displayLog = (host: any) => {
        const url = [`${host.protocol}://${host.address}`, typeof host.port === "number" && host.port].filter(Boolean).join(":") + (this.basePath === "/" ? "" : this.basePath);

        injector.logger.info(`WellKnown is available on ${url}/.well-known/openid-configuration`);
      };
```

packages/oidc-provider/src/OidcModule.ts
Romakita:
```typescript
private getBasePath() {
  const provider = this.injector.getProviders().find((provider) => provider.subType === INTERACTIONS);
  if (provider) {
     return provider.path.split('/interaction')[0] || "/"
  }  
  return "/"
}

$onRoutesInit() {
   this.basePath = this.getBasePath()
  //....
}
```

With this code, the basePath will always aligned with the InteractionsCtrl path :)

packages/oidc-provider/src/OidcModule.ts
Romakita: I'm not sure is this code is necessary. The OIDC client will consume the configured issuer url. So
if the server expose oidc under a subpath, the client should be configured with the correct url:

```
const { Issuer } = require('openid-client');
Issuer.discover('https://accounts.myserver.com/oidc') // => Promise
  .then(function (issuer) {
      
  });
```

For me the url rewrite isn't necessary because the .well-known will be available under "/oidc" if the InteractionsCtrl have a "/oidc" subpath.

packages/oidc-provider/test/oidc.integration.spec.ts
```diff
@@ -48,7 +48,7 @@ describe("OIDC", () => {
 
   it("should display the OIDC login page then login", async () => {
     const authRes = await request
-      .get("/auth?client_id=client_id&response_type=id_token&scope=openid+email&nonce=foobar&prompt=login&redirect_uri=http://localhost:3000")
+      .get("/oidc/auth?client_id=client_id&response_type=id_token&scope=openid+email&nonce=foobar&prompt=login&redirect_uri=http://localhost:3000")
```
Romakita, line 172: The unit shown that the changes are a breaking change :). I prefer to let the "/" subpath by default and let the developer to configure the InteractionsCtrl with the custom subpath !

Romakita (changes requested):

Romakita:
For the corevage, we will see that later. But we can add the unit test on the OIDCModule to cover correctly the code :)

packages/oidc-provider/src/OidcModule.ts
stefanvanherwijnen: > hum... let the user to configure the correct path from configuration could be a better options. Because today, /oidc isn't the default path and it will be a breaking change. no?

It is a breaking change, but the point is that the current implementation is incorrect in my opinion. Mounting oidc at / will make it impossible to use the TSed exception handler. For current users of oidc-provider and TSed this will break any manually configured clients, but if they use the .well-known route for endpoint discovery everything will still work (i.e. .well-known will point to the /oidc/* routes). For versioning this might be a problem, but I think it is a valid breaking change (or rather a 'fixing' change :grin: )

packages/oidc-provider/src/OidcModule.ts
stefanvanherwijnen: It is required according to the specifications of oidc (discovery document should be at the root path): https://github.com/panva/node-oidc-provider/blob/88fa711261401449292df5fa39d9349dd587120f/docs/README.md#to-a-koa-application

Romakita: Ok ^^ Well if the doc says so: p

Romakita:
Thanks, one more thing which does not seem to work: the Post() routes in the interactions do not seem to get registered when using koa. E.g. POST to /interactions/123/login results in a NOT_FOUND error with koa, but with express it results in SessionNotFound as expected. The main GET routes do work fine.

Any clue?

Hue... I haven't idea. I would have to go back to the code and reproduce the error to be able to answer you

stefanvanherwijnen:
> Thanks, one more thing which does not seem to work: the Post() routes in the interactions do not seem to get registered when using koa. E.g. POST to /interactions/123/login results in a NOT_FOUND error with koa, but with express it results in SessionNotFound as expected. The main GET routes do work fine.
> 
> Any clue?
> 
> Hue... I haven't idea. I would have to go back to the code and reproduce the error to be able to answer you

The problem seems to be that any Controller routes in any of the children of the InteractionController are not registered when using koa. I guess it is a problem somewhere in @tsed/platform-koa and route registration, but I can't find exactly where things go wrong.
If I don't use the `@Interaction` but register a controller directly with `@Controller` it does work, so I think it's somewhere in @tsed/oidc-provider not registering the routes properly in combination with koa. The strange thing is that they do show up in the server log upon boot, but result in NOT_FOUND.

Edit2:
Figured it out. For some reason `@Post("/login")` works in express but not in Koa. It should be `@Post("login")`

packages/oidc-provider/src/OidcModule.ts
stefanvanherwijnen: The basePath should not be the same as the interactions path. basePath (or /oidc) should point to the OIDC instance and the interactions should be on a separate route. I believe everything under /oidc is strictly specified by the OpenID Connect specification, and interactions are a user implementation.

Romakita:
> The problem seems to be that any Controller routes in any of the children of the InteractionController are not registered when using koa. I guess it is a problem somewhere in @tsed/platform-koa and route registration, but I can't find exactly where things go wrong.
> If I don't use the `@Interaction` but register a controller directly with `@Controller` it does work, so I think it's somewhere in @tsed/oidc-provider not registering the routes properly in combination with koa. The strange thing is that they do show up in the server log upon boot, but result in NOT_FOUND.

Haa ok. isn't normal. The custom interactions are mounted by oidc-provider from the configuration, by generating the policy options:
https://github.com/tsedio/tsed/blob/a22e24152523bcead7afdfefa63897a565ec9f90/packages/oidc-provider/src/services/OidcProvider.ts#L145

So, one of the explanation is oidc-provider have change something about custom interaction because, with express, the because with express it worked. With Koa, I don't know, but the principle should be the same. The policy provide an configuration abstraction...

I'll take a time to debug this issue ^^

stefanvanherwijnen:
I thought I fixed it, but `@Post(/login)` works in express but not koa, and `@Post(login)` works in koa but not express :sweat_smile: .
So I think it is a problem with [preceding slashes and koa](https://stackoverflow.com/questions/39666603/koa-router-issue-with-and-signin).

Edit: It is a problem with Controller mounting in koa:
```js
import {Controller, Get} from "@tsed/common";
import {EventCtrl} from "./EventCtrl";
@Controller({
  path: "/",
  children: [
    EventCtrl
  ]
})
export class RestCtrl {
}


@Controller("/events")
export class EventCtrl {
  @Get("/")
  get() {
  }
}

  mount: {
    "/rest": [
      `${rootDir}/controllers/**/*.ts`
    ],
  },
```
results in NOT_FOUND on /rest/events when using koa

Romakita:
Arf... so we have to manage the slash with koa for all routes added by decorator ?

Romakita:
ok because appending "/" can generate a route with double "//" in a specific case.

Romakita:
I work on a new PR to remove koa-router and express-router, so I 'll try to fix this issue in the same time.

stefanvanherwijnen:
Great, thanks.

Would you be OK with having a manually configured basepath which defaults to / now but to /oidc in the next major version? That way things won't break now but will be working correctly in the future.

Romakita:
The next major version isn't planed but yes sure :) You can add a note in the documentation page to inform the dev to the future breaking change.

packages/oidc-provider/src/OidcModule.ts
```diff
@@ -29,10 +32,12 @@ export class OidcModule {
   }
 
   async $onRoutesInit() {
-    if (this.platformName === "koa") {
-      this.app.use(rewrite("/.well-known/(.*)", "/oidc/.well-known/$1"));
-    } else {
-      this.app.use(expressRewrite("/.well-known/*", "/oidc/.well-known/$1"));
+    if (this.basePath !== "/") {
+      if (this.platformName === "koa") {
+        this.app.use(rewrite("/.well-known/(.*)", `${this.basePath}/.well-known/$1`));
```
Romakita: hum if this path is "/" this rewrite url will be "//.well-known/$1" I think it's problematic

Romakita (changes requested):

packages/oidc-provider/src/OidcModule.ts
stefanvanherwijnen: It only rewrites if basepath is not "/" 😉

Romakita: haaa yes sorry.

stefanvanherwijnen: I added some tests to make sure it works correctly  :smile:

Romakita (approved):

Romakita:
@stefanvanherwijnen I pushed new unit test to cover the code ;)

Romakita:
I let you to test the latest change before merging the PR ;) @stefanvanherwijnen

stefanvanherwijnen:
Alright thanks. I'll get to it later.

Romakita:
I can merge it if it’s ok for you

stefanvanherwijnen:
There are a still a few inconsistencies in the docs. Let me fix them first.

Romakita (approved):

stefanvanherwijnen:
Everything seems correct now. I also PRed the CLI plugin so if both are merged one can select OpenID Connect provider as a feature and everything should work :smile: .

Romakita:
Great Job @stefanvanherwijnen I'll merge in few minutes your awesome works ;)

PR #1563, opened 2021-10-05
Make onInit stateless by removing call to clear()

from-nibly:
## Information

Type | Bugfix
---|---
Fix | Yes

@Romakita looks like this solves the issue where Returns doesn't apply to classes. It also seems like this trips it up in other contexts. 

Not sure what else this will break. I guess we will see in the tests?


Romakita (approved):

Romakita:
@from-nibly thanks for the investigation and the fix ;)

PR #1595, opened 2021-10-21
fix(typeorm): fix cycling connections in typeorm when reinitializing

Oatelaus:
I've reported this issue in slack, you'd notice this when actually connecting to the database in tests - possibly if you cycle the TypeORM Module when running your application.

Calling `PlatformTest.create/reset` would go down into the `init` and `$onDestroy` functions. On the second test run, it would check `this.has(name)` and return the previous instance where the connections had been closed.


packages/orm/typeorm/src/services/TypeORMService.ts
```diff
@@ -23,6 +23,7 @@ export class TypeORMService {
     const name = getValue<string>(connectionOptions, "name", "default");
```
Romakita, line 24: I suggest to refactoring the createConnnection function to ensure that the connection is alway connected :)

```
export async function createConnection(connectionOptions: ConnectionOptions): Promise<Connection> {
  const connectionManager = getConnectionManager();
  const name = getValue<string>(connectionOptions, "name", "default");

  if (!connectionManager.has(name)) {
    const connection = connectionManager.create(connectionOptions!);
    await connection.connect();
  }

  const connection = connectionManager.get(name);

  return connection;
}

```

And in the typeormService:
```
  @Inject(InjectorService)
  private injector: InjectorService;

  /**
   *
   * @returns {Promise<"typeorm".Connection>}
   */
  async createConnection(connectionOptions: ConnectionOptions): Promise<any> {
    try {
      const connection = await createConnection(connectionOptions);
      this.injector.logger.info(`Connected with typeorm to database: ${connection.name}`);

      return connection;
    } catch (err) {
      /* istanbul ignore next */
      console.error(err);
      /* istanbul ignore next */
      process.exit();
    }
  }

```

Romakita (changes requested):

Romakita (approved):

Oatelaus:
I'll fix linting in a moment!

PR #1660, opened 2021-11-27
feat(mongoose): Add count option for VirtualRef

ochrstn:
## Information

Type | Breaking change
---|---
Feature | No

****

## Usage example

Adds support for the count option on VirtualRefs (https://mongoosejs.com/docs/populate.html#count):

```typescript
@Model()
class CompanyEntity {
  @VirtualRef({
      ref: 'UserEntity',
      localField: '_id',
      foreignField: 'membership.company',
      count: true,
      options: {
        match: { roles: 'client' },
      },
    })
  clientCount: number;
}
```

## Todos

- [x] ⚠️ How to generate the JsonSchema correctly so we get a number type there instead of the Ref (Array) Type?. I could use some help here 😅 
- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation

### DIFF
--- a/docs/tutorials/snippets/mongoose/virtual-references.ts
+++ b/docs/tutorials/snippets/mongoose/virtual-references.ts
@@ -22,6 +22,16 @@ class Band {
     options: {} // Query options, see http://bit.ly/mongoose-query-options
   })
   members: VirtualRefs<Person>;
+
+  @VirtualRef({
+    ref: Person, // The model to use
+    localField: "name",  // Find people where `localField`
+    foreignField: "band", // is equal to `foreignField`
+    // If `count` is true, 'memberCount' will be the number of documents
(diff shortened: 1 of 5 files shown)


Romakita (approved):

Romakita:
Hello @ochrstn 

Thanks for the PR :). for me it's ok.

Ok I see the VirtualRef set automatically the type. I think, in this case you have to test if the countOptions is given ^^

You can try that:
```typescript

  return useDecorators(
     StoreMerge(MONGOOSE_SCHEMA, schema), 
     type && !schema.count && (schema.justOne ? Property(type) : CollectionOf(type))
     schema.count && Proprety(Number)
);

```

See you
Romain

Romakita (approved):

PR #1668, opened 2021-12-01
feat(mongoose): Add an option to `Ref` for more accurate JsonSchema ...

ochrstn:
Add an option to `Ref` for more accurate JsonSchema based on `Groups` and populated Refs.

I would appreciate feedback on that 😅 

## Information

Type | Breaking change
---|---
Feature| No :pray:

****

## Usage example

```ts
@Model()
export class User {
    @ObjectID('id')
    _id: string;

    @Required()
    email: string;
}

@Model()
export class Task {
    @Property()
    title?: string;

    @Ref(User, {populatedGroups: ["pop"]})
    createdBy: Ref<User>;
}

@Controller('/tasks')
export class TaskController {

    constructor(@Inject(Task) private TaskModel: MongooseModel<Task>)

    @Get('/')
    @(Returns(200, Array).Of(Task).Groups("pop"))
    async list(): Promise<Task[]> {
      const tasks = await this.TaskModel.find({}).populate<{createdBy: User}>('createdBy')
      return tasks;

      // resulting schema has property "createdBy" limited to $ref type
    }
}
```

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [ ] Documentation

### DIFF
--- a/packages/orm/mongoose/src/decorators/ref.spec.ts
+++ b/packages/orm/mongoose/src/decorators/ref.spec.ts
@@ -211,4 +211,41 @@ describe("@Ref()", () => {
       });
     });
   });
+
+  describe("JsonSchema gets generated based on populated groups", () => {
+    class MyChildModel {
+      @Property()
+      test: string;
+    }
+
+    class MyParentModel {
+      @Property()
+      id: string;
+
+      @Ref(MyChildModel, {populatedGroups: ["group1", "group2"]})
+      child1: Ref<MyChildModel>;
+
+      @Ref(MyChildModel, {populatedGroups: ["group2"]})
+      child2: Ref<MyChildModel>;
+
+      @Ref(MyChildModel)
+      child3: Ref<MyChildModel>;
+    }
+
+    it("should reflect the populated groups options in the schema", () => {
+      const spec = getJsonSchema(MyParentModel, {
+        groups: ["group1", "group3"]
+      });
+
+      expect(spec.properties.child1.oneOf.length).toEqual(1);
+      expect(spec.properties.child1.oneOf[0].$ref).toEqual("#/definitions/MyChildModel");
+
+      expect(spec.properties.child2.oneOf.length).toEqual(1);
+      expect(spec.properties.child2.oneOf[0].type).toEqual("string");
+
+      expect(spec.properties.child3.oneOf.length).toEqual(2);
+      expect(spec.properties.child3.oneOf[0].type).toEqual("string");
+      expect(spec.properties.child3.oneOf[1].$ref).toEqual("#/definitions/MyChildModel");
+    });
+  });
 });
--- a/packages/orm/mongoose/src/decorators/ref.spec.ts
+++ b/packages/orm/mongoose/src/decorators/ref.spec.ts
@@ -243,13 +243,21 @@ describe("@Ref()", () => {
               }
             },
             type: "object"
+          },
+          MyChildModelGroup1Group3: {
+            properties: {
+              test: {
+                type: "string"
+              }
+            },
+            type: "object"
           }
         },
         properties: {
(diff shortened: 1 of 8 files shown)


packages/orm/mongoose/src/decorators/ref.spec.ts
```diff
@@ -211,4 +211,41 @@ describe("@Ref()", () => {
...
+    class MyChildModel {
+      @Property()
+      test: string;
+    }
+
+    class MyParentModel {
+      @Property()
+      id: string;
+
+      @Ref(MyChildModel, {populatedGroups: ["group1", "group2"]})
```
Romakita, line 222: I don't understand the goal of this option ? 
It's to solve the problem with the OneOf + Groups options ? If it's for that, We should fix the issue with Groups + OneOf/AllOf/etc... instead ?

In your, test I don't see the impact on the MyChildModel if the groups is given.

packages/specs/schema/src/utils/mapRefType.ts
```diff
@@ -0,0 +1,8 @@
+import {JsonSchema, JsonSchemaOptions} from "..";
```
Romakita: this import will create a Circular ref in the package ^^

Romakita (changes requested):

packages/orm/mongoose/src/decorators/ref.spec.ts
ochrstn: The goal is different here.

Without this option the schema type of `@Ref` property (child1, child2, child3 in the test) is always oneOf $ref and string.

Based on the presence of the `populatedGroups` option and matching groups the oneOf schema can be limited to be either $ref or string. This should of course correspond to the populated and not populated fields from a mongoose query that gets returned from an endpoint.

It does not and should not impact the MyChildModel. Only the properties of the MyParentModel.

Resulting schema of the MyParentModel:

child1: Because the endpoint group and ref group is matching I assume that the developer populated the child1 property in a query and returning it from the endpoint. The schema should be generated to reflect that it can't be a string anymore in that case.

child2: The opposite. Not populated property should be reflected in the schema to be only of type string.

child3: option not given, works live before

```
{
        "type": "object",
        "properties": {
            "id": {
                "type": "string"
            },
            "child1": {
                "oneOf": [
                    {
                        "$ref": "#/definitions/MyChildModel"
                    }
                ]
            },
            "child2": {
                "oneOf": [
                    {
                        "type": "string",
                        "examples": [
                            "5ce7ad3028890bd71749d477"
                        ],
                        "description": "Mongoose Ref ObjectId"
                    }
                ]
            },
            "child3": {
                "oneOf": [
                    {
                        "type": "string",
                        "examples": [
                            "5ce7ad3028890bd71749d477"
                        ],
                        "description": "Mongoose Ref ObjectId"
                    },
                    {
                        "$ref": "#/definitions/MyChildModel"
                    }
                ]
            }
        },
        "definitions": {
            "MyChildModel": {
                "type": "object",
                "properties": {
                    "test": {
                        "type": "string"
                    }
                }
            }
        }
    }
```

Romakita: ok I understand :)

I pushed small changes on your branch ;). I let you to review the code.

ochrstn: Thanks for the help. It looks amazing now

Romakita: Last point is about the coverage over @tsed/schema ^^

ochrstn: I will take a look 👀

Romakita:
Merged ;)

PR #1671, opened 2021-12-04
fix(schema): Fix oneOf/allOf/anyOf + Groups

Romakita:
## Information

- Isolate unit tests from common package 
- fix issue with OneOf + Groups


Romakita:
@ochrstn The last commit of this PR fix the issue with Groups + OneOf/AllOf/AnyOf :)

It should resolve the issue with Returns.OneOf

PR #1670, opened 2021-12-02
fix(mongoose): Fix json schema for `VirtualRef` if ref/type is given as string

ochrstn:
## Information

Type | Breaking change
---|---
Fix | No :pray:

****

## Usage example

see test case

## Todos

- [x] Tests
- [x] Coverage
- [ ] Example
- [ ] Documentation

PR #1672, opened 2021-12-04
fix(mongoose): Fix json schema for `VirtualRef` if MongooseModels entry not set yet

ochrstn:
## Information

Type | Breaking change
---|---
Fix | No

Sadly my last fix (https://github.com/tsedio/tsed/pull/1670) introduced a new bug.

`MongooseModels.get(ref)` can return undefined and the property doesn't even get serialized. With the fallback to `Object` it would behave like before.

That happens a lot in my project because

https://github.com/tsedio/tsed/blob/3e6ceced19a039200519c3882c39c4c4792c10de/packages/orm/mongoose/src/utils/createModel.ts#L11 gets executed after https://github.com/tsedio/tsed/blob/3e6ceced19a039200519c3882c39c4c4792c10de/packages/orm/mongoose/src/decorators/virtualRef.ts#L16

@Romakita any idea on how to fix the order problem?

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/orm/mongoose/src/decorators/virtualRef.spec.ts
+++ b/packages/orm/mongoose/src/decorators/virtualRef.spec.ts
@@ -2,7 +2,7 @@ import {Default, Format, getJsonSchema, getSpec, Post, Property, ReadOnly, Retur
 import {Store} from "@tsed/core";
 import {Model} from "@tsed/mongoose";
 import {MONGOOSE_SCHEMA} from "../constants";
-import {VirtualRef} from "./virtualRef";
+import {VirtualRef, VirtualRefs} from "./virtualRef";
 import {Controller} from "@tsed/di";
 import {BodyParams} from "@tsed/platform-params";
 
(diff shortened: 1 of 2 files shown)


Romakita:
@ochrstn  Maybe we can wrap the ref in arrow function instead ?

but the first step is a to add a unit or integration test that show the problem + expectation. Then we can find a solution.

See you
Romain

ochrstn:
> @ochrstn Maybe we can wrap the ref in arrow function instead ?
> 
> but the first step is a to add a unit or integration test that show the problem + expectation. Then we can find a solution.
> 
> See you Romain

I added a failing test.

I also tried to change the `VirtualRef` decorator so that it's more similar to the `Ref`

```ts
function getType(opts: any) {
  const ref = opts.ref || opts.type;
  return isString(ref) ? MongooseModels.get(ref) || Object : isArrowFn(ref) ? ref() : ref;
}

export function VirtualRef(options: string | MongooseVirtualRefOptions, foreignField?: string): Function {
  const opts = getInitialOpts(options, foreignField);
  const schema = mapToSchema(opts);
  return useDecorators(
    StoreMerge(MONGOOSE_SCHEMA, schema),
    JsonEntityFn((store) => {
      store.itemSchema.type(schema.count ? Number : lazyRef(getType(opts)));
      store.type = Object;
    })
  );
}
```

But that results `"$ref": "#/definitions/JsonSchema"` types for String references and other problems for Arrow function references

Romakita:
Fixed :)

PR #1677, opened 2021-12-07
Update SocketClientService to support custom paths

dimaKudr:
## Information
Problem: Default value for Socket.io server path is '/socket.io'. If you use some custom path for production code configuration, then your tests will fail as SocketClientService ignores that.
I've added path parameter to the get() so in tests, you could use the same path as for production code.
Without that parameter Socket Manager uses default path '/socket.io' and fails to connect.

Type | Breaking change
---|---
Fix | No

****
## Usage example

```typescript
const CUSTOM_WS_PATH = "/ws";

before(PlatformTest.bootstrap(Server, {
    platform: PlatformExpress,
    listen: true,
    httpPort: 8999,
    ...
    socketIO: { path: CUSTOM_WS_PATH }
}));

const service = PlatformTest.get<SocketClientService>(SocketClientService);
const client = await service.get("/test", CUSTOM_WS_PATH);

```
For more details see here: https://github.com/tsedio/tsed/blob/production/packages/third-parties/socketio/test/socket.integration.spec.ts

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [*] Documentation

### DIFF
--- a/packages/third-parties/socketio-testing/src/services/SocketClientService.ts
+++ b/packages/third-parties/socketio-testing/src/services/SocketClientService.ts
@@ -21,7 +21,7 @@ export class SocketClientService implements OnDestroy {
     const uri = `http://${address}:${port}${namespace}`;
 
     this.logger.info("Bind Socket.io client on:", uri);
-    const client = io(uri, path ? { path } : undefined);
+    const client = io(uri, path ? {path} : undefined);
 
     this.clients.set(namespace, client);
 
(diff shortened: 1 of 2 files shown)


Romakita (approved):

Romakita:
Hello @dimaKudr 
Thanks a lot for this improvement. There is a small issue with the linter ;). 

When it’s fixed I’ll merge and release this pr :)

see you 
Romain

Issue #1694, opened 2021-12-28
[FEAT] Add support for optional dependencies


derevnjuk:
## Informations

As a user, I would like to use the empty constructor `new Foo()`, to avoid registering `noop` dependencies if they are not required.

The new parameter decorator can be introduced to be used on constructor parameters, which marks the parameter as optional. At the same time, the IoC container should provide `null` or `undefined` if the dependency is not found.

```ts
interface Bar {
  // ...
}

const Bar: unique symbol = Symbol('Bar');

@Injectable()
class Foo {
  constructor(@Optional() @Inject(Bar) private readonly bar?: Bar) {}
}
```

Currently, while resolving an instance, the IoC container fails with the following error:
```
    TypeError:

   487 |   static getParamTypes(targetPrototype: any, propertyKey?: string | symbol): any[] {
   488 |     console.log(DESIGN_PARAM_TYPES, targetPrototype, propertyKey);
 > 489 |     return Reflect.getMetadata(DESIGN_PARAM_TYPES, targetPrototype, propertyKey!) || [];
       |                    ^
   490 |   }
```
However, there is a workaround. You can register a value provider manually like this:

```ts
registerProvider({provide: Bar, useValue: null});
```

Romakita:
Hello @derevnjuk 

Yes the DI isn’t designed to have optional dependencies.  I don’t know what is the exact usecase, because dependency always exist if it’s an injected class (tokenProvider===class). And circular dependency musn’t be injected like this (by constructor) and must use `@Inject` decorator for that (on property).

The error is a symptom of a circular dependencies, because type isn’t stored by typescript. 

Adding optional decorator is strange for me. In your example, the bar will never be injected… so what is the interest? That I don’t see ^^

I think isn’t a huge work to solve this but I want to be sure there is a real usecase.

See you ;)

derevnjuk:
> must use @Inject decorator for that (on property).

That's not exactly true. Even if you use a decorator on the property, it will raise the same exception.

> The error is a symptom of a circular dependencies, because type isn’t stored by typescript.

In this particular case, it has nothing to do with circular dependencies. Once you declare a dependency, `InjectorService` will try to resolve it using class provider (i.e `useClass`) by default.

```ts
deps = deps || Metadata.getParamTypes(provider.useClass);
```

Since you imply that a dependency is optional, you don't register a provider. As a result, `provider.useClass` is undefined. See the example below (inspired by `@tsed/mikro-orm` package):

[...]

PR #1697, opened 2021-12-29
feat(mikro-orm): provide the ability to set up a retry strategy

### LINKED ISSUE #1696: [FEAT] Provide the ability to set up a retry strategy
As a user, I would like to be able to implement and use a retry strategy on errors during a transaction, e.g automatic retry mechanism for optimistic locking.

## Informations

Once the `@Transactional` decorator is used, the interceptor should retry handling a request on the fails or for any other reason. The `@Transactional` should also provide the ability to enable/disable a retry strategy upon request.

So, we have to do the following things:

1. Declare an interface to allow implementing a custom retry strategy.
```ts
export interface RetryStrategy {
  acquire<T extends (...args: unknown[]) => unknown>(task: T): Promise<ReturnType<T>>;
}

export const RetryStrategy: unique symbol = Symbol("RetryStrategy");

```

2. Extend a `TransactionOptions` with a flag that indicates whether the automatic retry mechanism must be used.
```ts
export interface TransactionOptions {
  connectionName?: string;
  retry?: boolean;
}
```
3. Provide any documentation \ tutorials for that solution.

derevnjuk:
## Information

Type | Breaking change
---|---
Feature | No

## Usage example

To begin, you need to implement the `RetryStrategy` interface and use `registerProvider` to register it in the IoC container. Below you can find an example to handle occurred optimistic locks based on [an exponential backoff retry strategy](https://en.wikipedia.org/wiki/Exponential_backoff).  

`ExponentialBackoff` invokes passed function recursively is contained in a try/catch block. The method returns control to the interceptor if the call to the `task` function succeeds without throwing an exception. If the `task` method fails, the catch block examines the reason for the failure. If it's optimistic locking the code waits for a short delay before retrying the operation.

```typescript
import { OptimisticLockError } from '@mikro-orm/core';
import { RetryStrategy } from '@tsed/mikro-orm';

export interface ExponentialBackoffOptions {
  maxDepth: number;
}

export class ExponentialBackoff implements RetryStrategy {
  private depth = 0;

  constructor(private readonly options: ExponentialBackoffOptions) {}

  public async acquire<T extends (...args: unknown[]) => unknown>(
    task: T
  ): Promise<ReturnType<T>> {
    try {
      return (await task()) as ReturnType<T>;
    } catch (e) {
      if (
        this.shouldRetry(e as Error) &&
        this.depth < this.options.maxDepth
      ) {
        return this.retry(task);
      }

      throw e;
    }
  }

  private shouldRetry(error: Error): boolean {
    return error instanceof OptimisticLockError;
  }

  private async retry<T extends (...args: unknown[]) => unknown>(
    task: T
  ): Promise<ReturnType<T>> {
    await this.sleep(2 ** this.depth * 50);

    this.depth += 1;

    return this.acquire(task);
  }

  private sleep(milliseconds: number): Promise<void> {
    return new Promise(resolve => setTimeout(resolve, milliseconds));
  }
}

registerProvider({
  provide: RetryStrategy,
  useClass: ExponentialBackoff
});
```

Once it is done, you can enable an automatic retry mechanism using the `@Transactional` decoratory  like that:

```typescript
import {Controller, Post, BodyParams, Inject, Get} from "@tsed/common";
import {Transactional} from "@tsed/mikro-orm";

@Controller("/users")
export class UsersCtrl {
  @Inject()
  private usersService: UsersService;

  @Post("/")
  @Transactional({retry: true})
  create(@BodyParams() user: User): Promise<User> {
    return this.usersService.create(user);
  }

  @Get("/")
  getList(): Promise<User[]> {
    return this.usersService.find();
  }
}
```

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation

closes #1696
todo: the IoC container should provide `null` or `undefined` for the `RetryStrategy` interface by default until #1694 is closed

### DIFF
--- a/packages/orm/mikro-orm/readme.md
+++ b/packages/orm/mikro-orm/readme.md
@@ -287,7 +287,6 @@ export class UsersCtrl {
 }
 ```
 
-
 ## Contributors
 
 Please read [contributing guidelines here](https://tsed.io/CONTRIBUTING.html)
--- a/packages/orm/mikro-orm/readme.md
+++ b/packages/orm/mikro-orm/readme.md
@@ -200,6 +200,94 @@ export class UsersCtrl {
 }
 ```
 
+By default, the automatic retry policy is disabled. You can implement your own to match the business requirements and the nature of the failure. For some noncritical operations, it is better to fail as soon as possible rather than retry a coupe of times. For example, in an interactive web application, it is better to fail right after a smaller number of retries with only a short delay between retry attempts, and display a message to the user (for example, "please try again later").
+
+The `@Transactional()` decorator allows you to enable a retry policy for the particular resources. You just need to implement the `RetryStrategy` interface and use `registerProvider()` or `@OverrideProvider()` to register it in the IoC container. Below you can find an example to handle occurred optimistic locks based on [an exponential backoff retry strategy](https://en.wikipedia.org/wiki/Exponential_backoff).
+
+```ts
+import { OptimisticLockError } from '@mikro-orm/core';
+import { RetryStrategy } from '@tsed/mikro-orm';
+
+export interface ExponentialBackoffOptions {
+  maxDepth: number;
+}
+
+export class ExponentialBackoff implements RetryStrategy {
+  private depth = 0;
+
+  constructor(private readonly options: ExponentialBackoffOptions) {}
+
+  public async acquire<T extends (...args: unknown[]) => unknown>(
+    task: T
+  ): Promise<ReturnType<T>> {
+    try {
+      return (await task()) as ReturnType<T>;
+    } catch (e) {
+      if (
+        this.shouldRetry(e as Error) &&
+        this.depth < this.options.maxDepth
+      ) {
+        return this.retry(task);
+      }
+
+      throw e;
+    }
+  }
+
+  private shouldRetry(error: Error): boolean {
+    return error instanceof OptimisticLockError;
+  }
+
+  private async retry<T extends (...args: unknown[]) => unknown>(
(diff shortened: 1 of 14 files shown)


packages/orm/mikro-orm/readme.md
```diff
@@ -200,6 +200,94 @@ export class UsersCtrl {
 }
 ```
 
+By default, the automatic retry policy is disabled. You can implement your own to match the business requirements and the nature of the failure. For some noncritical operations, it is better to fail as soon as possible rather than retry a coupe of times. For example, in an interactive web application, it is better to fail right after a smaller number of retries with only a short delay between retry attempts, and display a message to the user (for example, "please try again later").
```
Romakita, line 203: You can add also this doc on /docs/tutorials/mikroorm.md ;)

Romakita (changes requested):

Romakita:
@derevnjuk Perfect. You just have to add the doc also on the /docs/tutorials/mikroorm.md and it's ok for merging this PR.

packages/orm/mikro-orm/readme.md
derevnjuk: Done, I hope I didn't miss anything)

Romakita: Perfect :)

Romakita (approved):

derevnjuk:
@Romakita oh, it seems you merged it is as-is. Let's stop GitHub workflow as soon as possible and then rebase commits)

PR #1700, opened 2021-12-30
ci: prevent `benchmarks` job for running on an action for fork

derevnjuk:
## Information

Type | Breaking change
---|---
Chore | No

****

See details in actions/github-script#42

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,6 +124,7 @@ jobs:
       - name: Run benchmarks
         run: yarn benchmarks
       - uses: actions/github-script@v4
+        if: ${{ github.event_name == 'pull_request' && !github.event.pull_request.head.repo.fork }}
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -104,6 +104,7 @@ jobs:
         run: yarn build
 
   benchmarks:
+    if: github.repository_owner == 'tsedio'
```
Romakita: haaaaa perfect ^^

Romakita (approved):

.github/workflows/build.yml
derevnjuk: Unfortunately, this doesn't work as expected) I will try to find out another solution. Github allows managing [permissions](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions) on job and workflow level, but I'm not sure there is a way to use them as an expression

PR #1748, opened 2022-01-30
build(deps): setup husky and lint-staged

derevnjuk:
## Information

Type | Breaking change
---|---
Chore | No

****

While working on #1747, I have encountered that husky and lint-staged does not work as expected. The main issue is in `.lintstagedrc`, `.prettierignore`, and `.eslintignore` config files:

- `.lintstagedrc` uses a wrong glob pattern to discover nested directories, ignoring other types of files (e.g. `.md`, `.json`, etc) in which we might be interested.
- `.prettierignore` ignores `docs` for some reason
- `.eslintignore` ignores both `docs` and `test` despite of the fact they can contain `.ts` as well


derevnjuk:
@Romakita I highly recommend you to remove ignorance for `docs` and `test` folders and perform linting and formatting for the whole project)

Romakita (approved):

Romakita:
@derevnjuk It seems to be ok for me ;)

derevnjuk:
@Romakita sorry for the late reply) I will do that in a separate PR then if don't mind

PR #1761, opened 2022-02-11
style(*): fix formatting issues

derevnjuk:
## Information

Type | Breaking change
---|---
Chore | No

****

To ensure code style consistency and provide a more convenient way to handle linting and formatting errors, `.eslintignore` and `.prettierignore` have been changed to cover tests and docs in addition to source code.


Romakita:
Hello @derevnjuk 
It's a great work. But... with the v7 branch it will be a nightmare to update it with these changes XD.
Thanks again for your help ;)
Romain

PR #1762, opened 2022-02-13
fix(json-mapper): fix serialization of additional properties

KODIKAS-NL:
Serialization of additional properties if @additionalProperties(true) decorator is set on Model

## Information

Type | Breaking change
---|---
Fix | No

****
I expected the addition of the @additionalProperties decorator to a Model to also allow serialization of any additional properties but it didn't. This changes fixes this.

### DIFF
--- a/packages/specs/json-mapper/src/utils/serialize.spec.ts
+++ b/packages/specs/json-mapper/src/utils/serialize.spec.ts
@@ -591,7 +591,7 @@ describe("serialize()", () => {
         }
       });
     });
-    it("should serialize model with additonal object props", () => {
+    it("should serialize model with additional object props", () => {
       @AdditionalProperties(true)
       class Model {
         @Property()
@@ -622,7 +622,7 @@ describe("serialize()", () => {
         }
       });
     });
-    it("should not serialize model with additonal object props", () => {
+    it("should not serialize model with additional object props", () => {
       class Model {
         @Property()
         id: string;
--- a/packages/specs/json-mapper/src/utils/serialize.spec.ts
+++ b/packages/specs/json-mapper/src/utils/serialize.spec.ts
@@ -1,5 +1,5 @@
 import {deserialize} from "@tsed/json-mapper";
-import {CollectionOf, Ignore, JsonHookContext, Name, Property} from "@tsed/schema";
+import {AdditionalProperties, CollectionOf, Ignore, JsonHookContext, Name, Property} from "@tsed/schema";
 import {expect} from "chai";
 import {parse} from "querystring";
 import {Post} from "../../test/helpers/Post";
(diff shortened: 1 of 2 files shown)


packages/specs/json-mapper/src/utils/serialize.ts
```diff
@@ -65,7 +65,10 @@ export function classToPlainObject(obj: any, options: JsonSerializerOptions<any,
 
   const entity = JsonEntityStore.from(type || obj);
 
-  return getSchemaProperties(entity, obj).reduce((newObj, [key, propStore]) => {
+  const additionalProperties = !!entity.schema.get("additionalProperties");
+  const schemaProperties = getSchemaProperties(entity, obj);
+
```
Romakita:
```typescript
const additionalProperties = !!entity.schema.get("additionalProperties");
const schemaProperties = getSchemaProperties(entity, obj);
const properties = new Set<string>()
const out: any = schemaProperties.reduce((newObj, [key, propStore]) => {
    properties.add(key);
    
    // ...
})

 if (additionalProperties) {
    objectKeys(obj).forEach((key: any) => {
       if (!properties.has(key)) {
          out[key] = obj[key];
       }
    });
  }
```

Romakita (changes requested):

packages/specs/json-mapper/src/utils/serialize.spec.ts
```diff
@@ -591,5 +591,63 @@ describe("serialize()", () => {
...
+
+      expect(result).to.deep.eq({
+        id: "id",
+        renamed: "myname",
+        additional: {
+          foo: "bar"
+        }
+      });
+    });
+    it("should not serialize model with additonal object props", () => {
```
Romakita: should not serialize model with **additional** object props

packages/specs/json-mapper/src/utils/serialize.spec.ts
```diff
@@ -591,5 +591,63 @@ describe("serialize()", () => {
         }
       });
     });
+    it("should serialize model with additonal object props", () => {
```
Romakita: should serialize model with **additional** object props

Romakita (changes requested):

Romakita:
Hello @KODIKAS-NL 
Thanks for this first PR ;)

Just minor changes! 
See you
Romain

packages/specs/json-mapper/src/utils/serialize.ts
KODIKAS-NL: This one gives me an error because the 'key' parameter in reduce is typed as 'string | number | symbol' :
image
I can solve this in one of 3 ways:
```suggestion
const properties = new Set<any>()
```
or
```suggestion
properties.add(key as string);
```
or
```suggestion
properties.add(key.toString());
```

Which do you prefer ?

packages/specs/json-mapper/src/utils/serialize.spec.ts
KODIKAS-NL: 😬

packages/specs/json-mapper/src/utils/serialize.ts
Romakita: properties.add(key as string);

is fine for me :)

Romakita (approved):

PR #1798, opened 2022-03-07
fix(di): get rid of `console.log()`

derevnjuk:
## Information

| Bug | No          |

---

While bootstrapping, `Interceptor` decorator calls `console.log()` multiple time with useless message:

```
=== [
  {},
  'handle',
  {
    value: [AsyncFunction: handle],
    writable: true,
    enumerable: false,
    configurable: true
  }
]
```

fixes #1796


Romakita:
Halalala….

Romakita:
Thanks ;)

PR #1814, opened 2022-03-19
Fix: a few possible errors in the PlatformHandler.onError method

### LINKED ISSUE #1812: [BUG] Possible bug in the PlatformHandler.onError method
## Information

- **Version:** 6.107.2
- Packages:
  @tsed/common

https://github.com/tsedio/tsed/blob/86ce929f40ad2b77e2646d7c909ff731e667b21e/packages/platform/common/src/services/PlatformHandler.ts#L164-L186

This method has a few issues:
- the `$ctx` can be possibly `undefined`: on line 166 `$ctx.data = er;` but on line 173 there is a check that `$ctx` is defined
- on line 179 `$ctx` can be undefined`
- on line 173 there is a check that `next` is not defined but on line 184 we have a check again that `next` is defined. So we can clean up line 184

AlexDede:
## Information

| Fix | No          |

---

Fixed: #1812

### DIFF
--- a/packages/platform/common/src/services/PlatformHandler.ts
+++ b/packages/platform/common/src/services/PlatformHandler.ts
@@ -163,26 +163,29 @@ export class PlatformHandler {
 
   protected async onError(er: Error, requestOptions: OnRequestOptions) {
     const {next, $ctx, metadata} = requestOptions;
-    $ctx.data = er;
+
+    if ($ctx) {
+      $ctx.data = er;
+    }
 
     if (!next) {
       throw er;
     }
 
     // istanbul ignore next
-    if (!$ctx || $ctx?.isDone()) {
+    if (!$ctx || $ctx.isDone()) {
       $log.warn({
         name: "HEADERS_SENT",
         message: `An error was caught after the headers were sent to the client. The error comes from the handler: ${metadata.toString()}`,
         stack: er.stack,
         origin: er,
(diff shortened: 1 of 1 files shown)


packages/platform/common/src/services/PlatformHandler.ts
```diff
@@ -163,26 +163,29 @@ export class PlatformHandler {
...
         origin: er,
-        request_id: $ctx.id
+        request_id: $ctx?.id
       });
 
       return;
     }
 
-    return next && next(er);
+    return next(er);
```
Romakita, line 188: next is not always given, it depend on how the handler have been built. Can you revert this change please ;)

Romakita (changes requested):

packages/platform/common/src/services/PlatformHandler.ts
AlexDede: This method throws the error if `next` is not defined a few lines above

AlexDede: So I can revert this line but in this case, I should remove

``` js
 if (!next) {
    throw er;
 }
```

But in this case, this can be a breaking change

Romakita: You’ve right.

Romakita (approved):

PR #1950, opened 2022-06-20
feat(mikro-orm): introduce middleware to isolate the request context

### LINKED ISSUE #1949: [FEAT] Use ALS to map EM to requests in the background
## Information

To utilize implicit transactions using the whole power of UoW, `@tsed/mikro-orm` provides the ability to register an interceptor using the `@Transactional` decorator. The interceptor is in charge of forking an entity manager within the async context mapping it to the particular call.

There are a few issues with this approach:

* Missed `@Transactional` might lead to race conditions or a growing memory footprint, that is why using a global entity manager is strictly forbidden by default in `mikro-orm`.
* It is almost impossible to propagate an entity manager to other services called in middlewares since they are called before an interceptor catch a request.

## Possible solution

We should introduce a middleware that comes in handy to map a fork to TsEd context. We should register this middleware as the last one just before request handlers and before any custom middleware that is using the ORM:

```ts
@Middleware()
export class MikroOrmContextMiddleware implements MiddlewareMethods {

  constructor(@Inject() private readonly registry: MikroOrmRegistry) {}

  public use(@Next() next: (...args: unknown[]) => void): void {
    const instances = [...this.registry.values()];
    const entityManagers = instances.map((orm) => orm.em);

    RequestContext.create(entityManagers, next);
  }
}
```

However, we can notice an issue using a util class from `mikro-orm`. Please take a look at the following example:
```ts
@Controller('/')
export class UserController {
  @Em() private readonly em!: EntityManager;

  @Get()
  @Use(MikroOrmContextMiddleware)
  public getAll(): Promise<void> {
    return this.em.find(User, {});
  }
}
```
Running the example above, `this.em.find()` will raise an error: `Using global EntityManager instance methods for context-specific actions is disallowed. If you need to work with the global instance's identity map, use the allowGlobalContext configuration option or fork() instead` (for details see  https://github.com/tsedio/tsed/issues/1289#issuecomment-800295504 ). To solve this issue Ts.ED provides the `@tsed/async-hook-context` package to manage natively the `async_hook`. 

So, we can rewrite the example above as follows:
```ts
@Middleware()
export class MikroOrmContextMiddleware implements MiddlewareMethods {

  constructor(@Inject() private readonly registry: MikroOrmRegistry) {}

  public use(@Next() next: (...args: unknown[]) => void): void {
    const instances = [...this.registry.values()];
    const entityManagers = instances.map((orm) => orm.em);
    getContext()?.set('entity_managers', entityManagers)
  }
}
```

derevnjuk:
## Information

| Feature | No          |

---

## Usage example

Please refer to #1949 for details. 

## Todos

- [x] Tests
- [x] Coverage
- [ ] ~Example~
- [ ] ~Documentation~

closes #1949


packages/orm/mikro-orm/package.json
```diff
@@ -37,9 +38,10 @@
   },
   "peerDependencies": {
     "@mikro-orm/core": ">=4.5.9",
+    "@tsed/async-hook-context": "^6.114.17",
```
Romakita: This package isn’t required anymore since the hook context is integrated to tsed/di

packages/orm/mikro-orm/src/services/MikroOrmEntityManagers.ts
```diff
@@ -0,0 +1,72 @@
...
+
+@Injectable()
+/**
+ * @internal Since 2022-06-20.
+ */
+export class MikroOrmEntityManagers {
+  private static readonly MIKRO_ORM_ENTITY_MANAGERS_KEY: unique symbol = Symbol("MIKRO_ORM_ENTITY_MANAGERS_KEY");
+  private readonly storage: AsyncLocalStorage<DIContext>;
+
+  private get context(): DIContext | undefined {
```
Romakita: The @InjectContext should do the job no?

packages/orm/mikro-orm/src/services/MikroOrmEntityManagers.ts
```diff
@@ -0,0 +1,72 @@
...
+    return this.storage.getStore();
+  }
+
+  constructor(@Inject() private readonly logger: Logger) {
+    if (AsyncLocalStorage) {
```
Romakita: Create storage is not necessary if your goal is just to retrieve context. InjectContext is here for that ^^

packages/orm/mikro-orm/src/services/MikroOrmEntityManagers.ts
```diff
@@ -0,0 +1,72 @@
...
+    } else {
+      throw new Error(`AsyncLocalStorage is not available for your Node.js version (${process.versions.node}).`);
+    }
+  }
+
+  public entries(): Map<string, EntityManager> | undefined {
+    return this.context?.get(MikroOrmEntityManagers.MIKRO_ORM_ENTITY_MANAGERS_KEY);
+  }
+
+  public set(em: EntityManager | EntityManager[]): this {
```
Romakita: [].concat(em) ?

packages/orm/mikro-orm/src/middlewares/MikroOrmContextMiddleware.ts
```diff
@@ -0,0 +1,17 @@
+import {Middleware, MiddlewareMethods} from "@tsed/platform-middlewares";
+import {Inject} from "@tsed/di";
+import {MikroOrmRegistry} from "../services/MikroOrmRegistry";
+import {Context} from "@tsed/common";
+import {MikroOrmEntityManagers} from "../services/MikroOrmEntityManagers";
+
+@Middleware()
+export class MikroOrmContextMiddleware implements MiddlewareMethods {
+  constructor(@Inject() private readonly store: MikroOrmEntityManagers, @Inject() private readonly registry: MikroOrmRegistry) {}
```
Romakita: Proposal: store => managers ?

Romakita (changes requested):
Hello @derevnjuk 
Nice PR. I just saw that this PR is draft but I already added some comments ;)

see you 
Romain

packages/orm/mikro-orm/src/services/MikroOrmEntityManagers.ts
derevnjuk: You are right. It should work well if we consider that it is used only with controllers. However, if we speak about Agenda (https://tsed.io/tutorials/agenda.html#create-a-new-service), the context will be null or undefined. I'm going to sort out how to handle both cases)

packages/orm/mikro-orm/src/services/MikroOrmEntityManagers.ts
derevnjuk: This implies casting `[]` to `EntityManger[]`, otherwise tsc raises compile-time error due to `ConcatArray`

```ts
const managers = ([] as EntityManager[]).concat(em);
```

packages/orm/mikro-orm/src/services/MikroOrmEntityManagers.ts
Romakita: The context is always initiate by the controller and especially by the DI. There is no other case that the context could be available or created by something else.

derevnjuk: This is true if we speak about `DIContext`. But in the case of orm, we should have an async context to isolate, handle and propagate transactions between different repositories and service methods. The current implementation doesn't allow us to do that, that is why PR is still in draft)

As shown in the example below, we expect that a fork will be used. `@UseRequestContext` acts as interception running `sendAdminStatistics` inside an async context and creating a fork in the background.

```ts
@Agenda({ namespace: "email" })
export class EmailJobService {
  @Orm()
  private readonly orm!: MikroOrm;

  @Every("60 minutes", {
    name: "maintenanceJob",
    /* ... and any option you would normally pass to agenda.every/define */ }
  })
  @UseRequestContext()
  async sendAdminStatistics(job: Job) {
    const admin = this.orm.em.findOne(User, {admin: true})

    // do something
  }
}
```
Once we replace the `@UseRequestContext` decorator with the current implementation of `@Transactional`, it won't work anymore due to the lack of `DIContext`. The previous one that used `DBContext` works well.

To understand the main issue, let's considered another example:

```ts
@Middleware()
export class MikroOrmContextMiddleware implements MiddlewareMethods {

  constructor(@Inject() private readonly registry: MikroOrmRegistry) {}

  public use(@Next() next: (...args: unknown[]) => void): void {
    const instances = [...this.registry.values()];
    const entityManagers = instances.map((orm) => orm.em);

    RequestContext.create(entityManagers, next);
  }
}

@Controller('/')
export class UserController {
  @Em() private readonly em!: EntityManager;

  @Get()
  @Use(MikroOrmContextMiddleware)
  public getAll(): Promise<void> {
    return this.em.find(User, {});
  }
}
```

Here we use the build-in `RequestContext` provided by `mikro-orm` to run the middleware sequence inside the same async context. However, it won't work as well since `PlatformHandler` is losing a context created by `RequestContext`. 

That means we should either sort out of how to use different instances of async local stores, or rework `PlatformHandler` to prevent losing context (then we can use build-in `RequestContext` from `mikro-orm`).

Romakita: Ok for the example with Agenda it make sense because agenda run code outside a RequestContext. RequestContext come from mikro orm right ? if it's from `@tsed/mikro-orm` the decorator name is confusing for me, because, in your example, it's a not request context (the request/response are missing) while the name seems to indicate that it is.

The name must be different, and should reflect the intent to create a context in order to execute the code in it.

> Here we use the build-in RequestContext provided by mikro-orm to run the middleware sequence inside the same async context. However, it won't work as well since PlatformHandler is losing a context created by RequestContext.

Ok I understand. It make sense if you want to create multiple context and isolated transaction, independently from each other.

Romakita (approved):

packages/orm/mikro-orm/src/MikroOrmModule.ts
```diff
@@ -18,24 +21,45 @@ declare global {
...
+    return this.registry.clear();
+  }
+
+  public $beforeRoutesInit(): void {
+    /**
+     * Use a raw middleware to create an async context,
+     * since {@link PlatformHandler} is losing a context created by {@link RequestContext}.
+     * For details see https://github.com/tsedio/tsed/issues/1289
+     */
+    this.app.use((req: Req, res: Res, next: Next) => this.createContext(next));
```
Romakita: This instruction won't work with Koa.

I think we need to find another way to wrap the context correctly.

Romakita: I have to rework the PlatformHandler to run the next method inside a context right ? or give you way to wrap each handler in a context.

Romakita (changes requested):

packages/orm/mikro-orm/src/MikroOrmModule.ts
Romakita: With this https://github.com/tsedio/tsed/pull/1957
You'll able to wrap the next function with the appropriate hook :)

```typescript
$alterRunInContext(next, $ctx) {
  return this.createContext(next)
}
```

derevnjuk: We can easily overload this middleware to cover both cases:

```ts
interface RawMiddleware {
  (req: Express.Request, res: Express.Response, next: Express.NextFunction): unknown;

  (context: Koa.Context, next: Koa.Next): unknown;
}

const rawMiddleware: RawMiddleware = (reqOrContext: Express.Request | Koa.Context, resOrNext: Express.Response | Koa.Next, next?: Express.NextFunction) =>
  this.createContext(next ?? resOrNext);

this.app.use(rawMiddleware);
```

This is not the best solution but should work.

> I have to rework the PlatformHandler to run the next method inside a context 

This may solve the initial issue. If it is possible to do with minimum effort, I would prefer to follow this solution.

> You'll able to wrap the next function with the appropriate hook :)

I have already tried to do the same) Unfortunately, it didn't work. Anyway, let's me check one more time

derevnjuk: @Romakita as soon as we rework `runInContext` as follows:

```ts
runInContext(next: Function) {
  return runInContext(this, () => this.injector.alterAsync("$alterRunInContext", next, this));
}
```

It should work properly. So, I guess we can confidently proceed with this approach

Romakita: Hum, not sure if it's correct... next won't be called in this case

Romakita: here is the right code:

```typescript
    return runInContext(this, () => {
      next = this.injector.alter("$alterRunInContext", next, this)
      return next()
    });
```

And in your module, it should be written as follow:
```typescript
$alterRunInContext(next, $ctx) {
  return () => this.createContext(next)
}
```

Romakita: Seems to be ready :
https://github.com/tsedio/tsed/pull/1957 :)

Romakita: #1957 Merged :)

Romakita (approved):

Romakita:
PR seems to be ok for me @derevnjuk . Can I merge it ? Documentation is up to date ?

See you
Romain

Romakita:
maybe it's a time to finish this doc on mikro-orm, now that the module is on mikro-orm v5?

https://github.com/mikro-orm/mikro-orm/pull/2587

derevnjuk:
@Romakita documentation is up to date, you can merge PR without any issues

PR #1957, opened 2022-06-24
feat(di): provide the ability to bind an async context

Romakita:


packages/di/src/domain/DIContext.ts
```diff
@@ -92,6 +93,13 @@ export class DIContext extends Map<any, any> implements ContextMethods {
   async emit(eventName: string, ...args: any[]) {
     return this.injector?.emit(eventName, ...args);
   }
+
+  runInContext(next: Function) {
+    return runInContext(this, () => {
+      next = this.injector?.alter("$alterRunInContext", next, this);
+      return next();
```
derevnjuk: This might lead to NPE since `this.injector` can be null or undefined
```suggestion
      next = this.injector?.alter("$alterRunInContext", next, this) ?? next;
      return next();
```

packages/di/src/domain/DIContext.ts
```diff
@@ -92,6 +93,13 @@ export class DIContext extends Map<any, any> implements ContextMethods {
   async emit(eventName: string, ...args: any[]) {
     return this.injector?.emit(eventName, ...args);
   }
+
+  runInContext(next: Function) {
+    return runInContext(this, () => {
+      next = this.injector?.alter("$alterRunInContext", next, this);
```
derevnjuk: I suggest providing the ability to use `async` hook. It might be useful in some cases.

```suggestion
    return runInContext(this, async () => {
      next = await this.injector?.alterAsync("$alterRunInContext", next, this);
```

derevnjuk (commented):
I guess we should introduce a new interface to describe the contract for `$alterRunInContext`, e.g.

```ts
export interface AlterRunInContext {
    $alterRunInContext(next: (...args: unknown) => unknown, ctx: DIContext): () => unknown | Promise<() => unknown>;
}
```

packages/di/src/domain/DIContext.ts
Romakita: well seen :)

Romakita: done  :D

derevnjuk (approved):

PR #1986, opened 2022-07-18
feat(prisma): upgrade to V4

tomapp:
## Information

| Chore | Yes          |

---

I think it require a bit further investigations actually. But according to prisma there's not much changes to deal with.

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/orm/prisma/package.json
+++ b/packages/orm/prisma/package.json
@@ -24,25 +24,25 @@
   },
   "private": false,
   "dependencies": {
-    "@prisma/generator-helper": "^3.10.0",
-    "@prisma/sdk": "^3.10.0",
+    "@prisma/generator-helper": "^4.0.0",
+    "@prisma/sdk": "^4.0.0",
     "change-case": "^4.1.2",
     "pluralize": "^8.0.0",
     "ts-morph": "^12.0.0",
     "tslib": "2.4.0"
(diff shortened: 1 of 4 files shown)


Romakita (approved):

packages/orm/prisma/package.json
```diff
@@ -24,25 +24,25 @@
...
     "@tsed/json-mapper": "6.117.2",
     "@tsed/schema": "6.117.2",
     "@types/change-case": "^2.3.1",
     "@types/pluralize": "0.0.29",
-    "prisma": "^3.10.0"
+    "prisma": "^4.0.0"
   },
   "peerDependencies": {
-    "@prisma/client": "^2 || ^3",
+    "@prisma/client": "4.0.0",
```
martin31821, line 45: Hey, since prisma 4.1.1 is out these days, this gives an error. I think this value should be ^4.0.0?

Romakita: Yes, I missed this wrong peerDependencies declaration during the code review. can you make a short PR @martin31821 or @tomapp  please ?

Romakita: it should be >=4.0.0

PR #1997, opened 2022-07-21
feat(mikro-orm): provide the ability to avoid an explicit transaction

### LINKED ISSUE #1996: [FEAT] Provide the ability to avoid an explicit transaction within the context
## Information

As a user, I would like to have the ability to use the implicit transaction or refuse them in general within the context:

```ts
const user = await em.findOne(User, 1);
user.email = 'foo@bar.com';
const car = new Car();
user.cars.add(car);

await em.persistAndFlush(user);
```

We can provide the ability to disable the usage of explicit transactions making sure that users can control this behavior via the decorator's options.

The new options can be introduced, extending the `TransactionOptions` as follows:

```ts
export interface TransactionOptions {
  retry?: boolean;
  isolationLevel?: IsolationLevel;
  contextName?: string;
  enable?: boolean;
  flushMode: FlushModel
}
```

derevnjuk:
## Information

| Feature | No          |

---

In some cases, you might need to avoid an explicit transaction, but preserve an async context to prevent the usage of the global identity map. For example, starting with v3.4, the MongoDB driver supports transactions. Yet, you have to use a replica set, otherwise, the driver will raise an exception.

## Usage example

To prevent `@Transactional()` use of an explicit transaction, you just need to set the `disabled` field to `true`:

```typescript
@Post("/")
@Transactional({disabled: true})
create(@BodyParams() user: User): Promise<User> {
  return this.usersService.create(user);
}
```

You can also set the [flushing strategy](https://mikro-orm.io/docs/unit-of-work#flush-modes) for the transaction by setting the `flushMode`:

```typescript
@Post("/")
@Transactional({flushMode: FlushMode.AUTO})
create(@BodyParams() user: User): Promise<User> {
  return this.usersService.create(user);
}
```

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation

closes #1996

### DIFF
--- a/docs/tutorials/mikroorm.md
+++ b/docs/tutorials/mikroorm.md
@@ -112,7 +112,7 @@ export class UsersService {
 
 It's also possible to inject an ORM by its context name:
 
-```ts
+```typescript
 import {Injectable} from "@tsed/di";
 
 @Injectable()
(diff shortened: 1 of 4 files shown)


Romakita (approved):

Romakita:
Perfect :) @derevnjuk 

You can click on merge button when you are ready ;)

See you :D

PR #2023, opened 2022-08-05
Fix peerDep of @prisma/client to match dependency

martin31821:
## Information

| Chore | No          |

### DIFF
--- a/packages/orm/prisma/package.json
+++ b/packages/orm/prisma/package.json
@@ -43,7 +43,7 @@
     "prisma": "^4.0.0"
   },
   "peerDependencies": {
-    "@prisma/client": "^4.0.0",
+    "@prisma/client": ">=4.0.0",
     "@tsed/core": "^6.125.1",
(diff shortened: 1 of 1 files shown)


packages/orm/prisma/package.json
```diff
@@ -43,7 +43,7 @@
     "prisma": "^4.0.0"
   },
   "peerDependencies": {
-    "@prisma/client": "4.0.0",
+    "@prisma/client": "^4.0.0",
```
Romakita: It should be >=4.0.0

Romakita (changes requested):

packages/orm/prisma/package.json
martin31821: Should I change all deps to >=4.0.0 then?

Romakita: no just for the peerDependencies, to allow future major version.

Romakita (approved):

PR #2058, opened 2022-08-25
fix(schema): adjust `examples` schema value placement  when `@Exampe` used in conjunction with `@CollectionOf`

ostridm:
## Information

| Fix | No          |

---

## Usage example

```typescript
import {} from "@tsed/schema";

class Model {
  @Example("Examples")
  @CollectionOf(String)
  collection: string[];
}
```
expected to produce schema

```typescript
{
  properties: {
    collection: {
      items: {
        examples: ["Examples"],
        type: "string"
      },
      type: "array"
    }
  },
  type: "object"
}
```

### DIFF
--- a/packages/specs/schema/src/decorators/common/example.spec.ts
+++ b/packages/specs/schema/src/decorators/common/example.spec.ts
@@ -1,5 +1,5 @@
 import {descriptorOf, useDecorators} from "@tsed/core";
-import {Description, getSpec, In, Name, OperationPath, Path, Pattern, SpecTypes} from "@tsed/schema";
+import {CollectionOf, Description, getSpec, In, Name, OperationPath, Path, Pattern, SpecTypes} from "@tsed/schema";
 import {getJsonSchema} from "../../utils/getJsonSchema";
 import {Example} from "./example";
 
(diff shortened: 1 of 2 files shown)


derevnjuk (changes requested):
This is breaking change. For example, let's consider the following dto:

```ts
class Model {
  @Example(["Example"])
  @CollectionOf(String)
  collection: string[];
}
```

After introducing these changes, the model above will produce totally different schema from what users expect:

```js
{
  properties: {
    collection: {
      items: {
        examples: [["Examples"]],
        type: "string"
      },
      type: "array"
    }
  },
  type: "object"
}
```
Instead of the expected one:

```js
{
  properties: {
    collection: {
      items: {
        examples: ["Examples"],
        type: "string"
      },
      type: "array"
    }
  },
  type: "object"
}
```
or

```js
{
  properties: {
    collection: {
      items: {
        type: "string"
      },
      examples: [["Examples"]],
      type: "array"
    }
  },
  type: "object"
}
```
Moreover, it also implies violating LSP and postcondition. IMO I would avoid merging this PR

Romakita:
@derevnjuk Yes, totally agree. I think it’s just an issue because the decorator isn’t used correctly when the prop is a collection of a primitive type.

Romakita:
@ostridm I’m sorry, I suggested to make a PR for nothing. The correct solution was already mentionned in the issue and by @derevnjuk.

thanks for your investment ;)
See you

PR #2078, opened 2022-09-03
feat(objection): add model relationship decorators

rossyman:
## Information

| Type      | Breaking change |
| ------------------| ------- |
| Feature | No          |

First time contributor here 😄  If there are any striking concerns about this PR then please let me know, as I'm happy to work on it. I thought that it'd be useful for developers to use decorators to describe relationships via Objection.

This PR _hopefully_ honours the way that Objection's Graph and Relationship Querying APIs work -- whilst also ensuring the consumer's domain model is a bit more strict, by actually defining related properties on an entity as optional, instead of only loosely defining them via `relationMappings`.

The _sensible_ defaults these decorators provide for joining keys are based on the use of `@IdColumn` and `@Entity`. The defaults supplied are based on the [Objection relationship documentation](https://vincit.github.io/objection.js/guide/relations.html) and follow the same nomenclature with the ability for a user to override the keys / table path in the decorator factory params.

---

## Usage example

```typescript
/**
 * All work in a similar manner:
 * - @HasMany, @HasOne, @HasOneThroughRelation, @ManyToMany, @RelatesTo
 */
import { Entity, IdColumn, BelongsToOne } from '@tsed/objection';

@Entity("user")
class User extends Model {
  @IdColumn()
  id!: string;
}

@Entity("movie")
class Movie extends Model {

  @IdColumn()
  id!: string;

  ownerId!: string;

  @BelongsToOne()
  owner?: User;
}

// Retrieve the related user
const owner = await Movie.relatedQuery('owner').for(1);

// Retrieve the movie with their owner
const movie = await Movie.query().withGraphFetched('owner');
```

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation


rossyman:
Some extra thought I had on this PR:
- [x] ~It would be nice if we could infer the `modelClass` from the type defined on the property (Not even sure if this is possible)~
- [x] ~It would be nice if we could make the join options supplied to this decorator less nested~

rossyman:
Last thought I had on this, was that:
- [x] ~It would be nice if we could set a _sensible_ default for relationship joins~

packages/orm/objection/src/decorators/relatesTo.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    class User extends Model {
+      @IdColumn()
+      id!: string;
+    }
+    @Entity("movie")
+    class Movie extends Model {
+      @IdColumn()
+      id!: string;
+      userId!: string;
+      @RelatesTo(Model.BelongsToOneRelation)
```
Romakita, line 17: Not sure to understand the differentes between BelongsToOne and this decorator.

Why need to give `Model.BelongsToOneRelation` as parameter ?

It could be default value for the first parameter ?

Romakita: Ok RelatesTo is a technical decorator for others decorators. ^^

Romakita (approved):

Romakita:
Hello @rossyman,

It seems to be ok for me :) Good PR. 

The documentation over the new decorators is missing. Don't forget to add it in /docs/tutorials/objection.md ;)

See you
Romain

rossyman:
Updated the documentation and fixed the build/test phase -- One point, is that I'm not sure what I need to do for `viewpress` to find these new decorators, as at the moment, they're not showing up in the `<ApiList>` component in the markdown.

Romakita:
@rossyman you can take as example the mongoose.md where the ApiList is used to display mongoose decorators ;)

docs/tutorials/objection.md
```diff
@@ -78,6 +78,83 @@ export class User extends Model {
...
+```
+
+### Default joining keys
+
+When used in conjunction with @@Entity@@ and @@IdColumn@@, Ts.ED attempts to provide you with a sensible default for your join keys out of the box, reducing the amount of boilerplate you need to write.
+
+In the instance of @@BelongsToOne@@, the default join keys will be:
+
+```json
+// (e.g.: `movie.ownerId` and `user.id`)
```
Romakita: This comment may be a potential error in a json fragment code. 

You can use a block tip after code example:

```
::: tip

my tip

:::
```

rossyman:
Ah okay so it seems the API is queried from `tsed.io` -- So makes sense these don't show up yet as they're not released.

Romakita:
> Ah okay so it seems the API is queried from tsed.io -- So makes sense these don't show up yet as they're not released.

It's possible to regenerate locally the api.json file by running the command: `yarn docs:serve`. But this command, works only on linux/mac Os.

I'll checkout your PR to verify if the ApiList is correctly configured for Objection page.

See you
Romain

rossyman:
Tests are running into an issue with array types. Sadly this issue (https://github.com/microsoft/TypeScript/issues/7169) has been open for 6 years now, so I'll add in some extra parameters to the configuration for relationships that are plural to expose this type.

rossyman:
I can now confirm that the tests and docs are all building / passing correctly 😄 

Screenshot 2022-09-06 at 11 33 17

Screenshot 2022-09-06 at 11 25 13

Romakita (changes requested):

packages/orm/objection/src/decorators/hasMany.spec.ts
```diff
@@ -14,7 +14,7 @@ describe("@HasMany", () => {
     class User extends Model {
       @IdColumn()
       id!: string;
-      @HasMany()
+      @HasMany({modelClass: Pet})
```
Romakita: I prefer that because the type is mandatory:

```typescript
@HasMany(Pet) // without circular ref
@HasMany(() => Pet) // for circular ref
pets?: Pet[];
```

Keep in mind, the possibility to support the circular ref. So the decorator should support this implementation:

```typescript
@HasMany(() => Pet) // Note: it's already supported by CollectionOf
pets?: Pet[];
```

Potential circular ref, that we need to support:

```typescript
class User {
  role: Role
}

class Role {
   users: User[]
}
```

packages/orm/objection/src/decorators/manyToMany.spec.ts
```diff
@@ -47,9 +47,10 @@ describe("@ManyToMany", () => {
       @ManyToMany({
```
Romakita: Idem.

```typescript
@ManyToMany(Pet)
@ManyToMany(() => Pet)

@ManyToMany(Pet, {...})
@ManyToMany(() => Pet, {...})
```

packages/orm/objection/src/decorators/relatesTo.ts
```diff
@@ -16,6 +16,7 @@ import {createRelationshipMapping} from "../utils/createRelationshipMapping";
 export function RelatesTo(relation: RelationType, opts?: RelationshipOpts): PropertyDecorator {
   return useDecorators(
```
Romakita, line 17: ho ok. You already implemented the collectionOf :) nice.

You can simplify you code. The useDecorators filters decorators when the value is falsy ;):

```typescript
return useDecorators(
    opts?.type ? CollectionOf(opts.type) : Property(),
    StoreFn((store, params) => {
      if (isFunction(opts?.type)) {
         opts.type = opts.type()  // resolve the circular ref
      } 
      store.set(OBJECTION_RELATIONSHIP_KEY, createRelationshipMapping(params, relation, opts));
    })
    
 ```   
 Note: I prefer `type` rather than `modelClass` because, type is already used by the @tsed/core @tsed/schema to define the property type.

File: packages/orm/objection/src/domain/RelationshipOpts.ts [hunk 6]
```diff
@@ -1,5 +1,12 @@
 import {RelationJoin, RelationMapping} from "objection";
 
-export type RelationshipOpts = Partial<RelationJoin> & Omit<RelationMapping<any>, "relation" | "join" | "modelClass">;
+import {Type} from "@tsed/core";
 
-export type RelationshipOptsWithoutThrough = Omit<RelationshipOpts, "through">;
+export type PluralRelationshipOptsWithThrough = Partial<RelationJoin> &
+  Omit<RelationMapping<any>, "relation" | "join" | "modelClass"> & {modelClass: Type<any>};
```
Romakita: modelClass to type

packages/orm/objection/src/utils/createJoinKeys.ts
```diff
@@ -22,7 +22,7 @@ export function createJoinKeys(targetClass: any, target: any, propertyKey: strin
     case Model.HasOneThroughRelation:
       return {
         from: SOURCE_MODEL_KEY,
-        through: opts?.through || {
+        through: (opts && "through" in opts && opts.through) || {
```
Romakita: why have you changed this line ?

Romakita: for me `opts?.through` is equivalent to `opts && "through" in opts && opts.through`

packages/orm/objection/src/utils/createJoinKeys.ts
```diff
@@ -1,14 +1,14 @@
...
+import {Metadata} from "@tsed/core";
 import {RelationshipOpts} from "../domain/RelationshipOpts";
 
 /**
  * @ignore
  */
 export function createJoinKeys(targetClass: any, target: any, propertyKey: string, relation: RelationType, opts?: RelationshipOpts) {
-  const FOREIGN_MODEL = Metadata.getType(target, propertyKey);
-  const SOURCE_MODEL_KEY = `${targetClass.tableName}.${opts?.from || target.idColumn}`;
+  const FOREIGN_MODEL = (opts && "modelClass" in opts && opts.modelClass) || Metadata.getType(target, propertyKey);
```
Romakita: opts?.type || Metadata.getType(target, propertyKey)

Romakita: hum the usage of targetClass and target in the same function is confusing ^^.

```typescript
createJoinKeys(targetClass: any, foreignClass: any, propertyKey: string, relation: RelationType, opts?: RelationshipOpts) {
```

Romakita (changes requested):

Romakita:
@rossyman Great job ;)

Just small cosmetics changes !

See you
Romain

packages/orm/objection/src/utils/createJoinKeys.ts
rossyman: Since `RelationshipOpts` is now a union, we need to discriminate on that otherwise TypeScript will always think `through` does not exist. Adding in a type-guard here instead to make it a bit easier to read.

packages/orm/objection/src/decorators/relatesTo.ts
rossyman: Luckily objection can also resolve circular references in the same way out of the box, so there's no need to re-assign the value of `type` to the called function as Objection will do that for us 😄

rossyman: Adding in an extra check here to ensure `.prototype` isn't in `opts?.type`, as `isFunction` returns true for both Constructable and Callable types (So this code would execute on both the Model and factory function)

packages/orm/objection/src/utils/createJoinKeys.ts
rossyman: So the `propertyKey` metadata is installed on the `Model` class, so that's why we need to differentiate across `targetClass` and `target` -- `tableName` is on the `targetClass` instead of `target` since it's defined through a class decorator.

rossyman: Renamed to `targetClass` and `targetSuper`, hopefully a bit easier to understand now.

rossyman:
Good to go! 🚀 

Screenshot 2022-09-08 at 00 42 17

rossyman:
~Just on a side note, `Objection.js` isn't listed under the Plugins sidebar section on the docs site.~ -- Added it in this PR as a separate commit.

Romakita:
@rossyman Perfect PR ;)

Ready to merge for me!

Romakita (approved):

PR #2080, opened 2022-09-06
feat(schema): add deprecated decorator for params and props

RobertoGeuke:
## Information

| Feature | No          |

---
**Params**
```typescript
import {Controller, Get, QueryParams} from '@tsed/common';
import {Deprecated} from "@tsed/schema";

@Controller('/model')
class ModelController {
    @Get()
    handler(@QueryParams('filter') @Deprecated() filter: string) {
        // ...
    }
}
```
Will produce the following OAS
```
"/model": {
    "get": {
        "parameters": [
            {
                "deprecated": true,
                "in": "query",
                "name": "filter",
                "required": false,
                "schema": {
                    "type": "string"
                }
            },
        ]
    },
}
```

**Props**
```typescript
import {Deprecated} from "@tsed/schema";
class Model {
    @Deprecated()
    title: string;
}
```
Will produce the following OAS
```
"Model": {
    "type": "object",
    "properties": {
         "newsTitle": {
             "type": "string",
             "deprecated": true
         }
    }
}    
```

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/specs/schema/src/decorators/operations/deprecated.spec.ts
+++ b/packages/specs/schema/src/decorators/operations/deprecated.spec.ts
@@ -1,4 +1,5 @@
-import {Deprecated, getSpec, OperationPath, SpecTypes} from "@tsed/schema";
+import {Deprecated, getSpec, OperationPath, Returns, SpecTypes} from "@tsed/schema";
+import {QueryParams} from "@tsed/platform-params";
 
 describe("Deprecated", () => {
   it("should store metadata (swagger)", () => {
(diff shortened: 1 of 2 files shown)


Romakita:
Ok. Here it's because the `default` case in the `switch` isn't tested.

Adding this test should resolve your issue:
```typescript
it('should throw an error when the decorator is as class decorator', () => {
       class Model {

       }
       const error = catchError(() => {
           Deprecated(true)(Model)    
       })
       
       expect(error?.message).toEqual("")
)
```
See you
Romain

Romakita (approved):

PR #2131, opened 2022-10-20
fix(platform-params): improve documentation around validator/json-mapper when input param decorator is used

### LINKED ISSUE #2130: Required validation on parameters is not enforced unless UseValidation used
## Information

- **Version:** 6.114.18
- Packages:

I'm using `@Required()` in front of couple of parameters in my controller, but that is not effective until added `@UseValidation` decorator as well.
Is it supposed to be so? If yes - then all of the documentation should be updated.
I'd actually expect `@Required()` alone to be sufficient.

## Example

```ts
import { Controller, Get, HeaderParams, UseValidation } from '@tsed/common';
import { Required, Returns } from '@tsed/schema';

@Controller('/test')
export class TestCtrl {

	@Get()
	@Returns(200)
	public test(
		@UseValidation() @Required() @HeaderParams('x-a') a: string,
		@UseValidation() @Required() @HeaderParams('x-b') b: number
	) {
		return {};
	}
}
```

## Acceptance criteria

- [ ] Throw 400 without `@UseValidation` decoration

Right now, if not using that, it passes 200 OK.

Romakita:
Closes: #2130

### DIFF
--- a/docs/docs/controllers.md
+++ b/docs/docs/controllers.md
@@ -12,10 +12,12 @@ Controllers are responsible for handling incoming **requests** and returning **r
 
 <figure><img src="./../assets/client-controllers.png" style="max-height: 300px; padding: 10px; background: white"></figure>
 
-A controller is here to handle a specific request for a given HTTP verb and Route. The routing service is responsible for
+A controller is here to handle a specific request for a given HTTP verb and Route. The routing service is responsible
+for
 managing and dispatching request to the right Controller.
 
-In order to create a basic controller, we use classes and **decorators**. Decorators associate classes with required metadata and enable Ts.ED to create a routing map.
+In order to create a basic controller, we use classes and **decorators**. Decorators associate classes with required
+metadata and enable Ts.ED to create a routing map.
 
 ## Routing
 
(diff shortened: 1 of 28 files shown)


docs/docs/controllers.md
```diff
@@ -105,14 +113,32 @@ In order to avoid such side-effects, simply move `findAll()` method above `findO
...
-- @@RawPathParams@@: `Express.request.params` without transformation and validation,
-- @@QueryParams@@: `Express.request.query`
-- @@RawQueryParams@@: `Express.request.query` without transformation and validation,
+::: tip
+
+Some decorators run automatically `validator` and `json-mapper`. This behavior can be changed (if the decorator is flagged as configurable)
+like this:
+
+```typescript
+@HeaderParams("x-header", { useValidator: true, useMapper: true })
```
gullerya: At least for the version 6.100... the syntax is a bit different: `@HeaderParams({ expression: 'x-header', useValidation: true })`.
I've tried out the one as in your example and it is not playing well from the Typescript perspective, FYI.

Romakita: You've right I'll fix that :)

PR #2151, opened 2022-11-07
2145 update cache manager to latest to support refreshthreshold

### LINKED ISSUE #2145: Update cache-manager to latest to support refreshThreshold
## Information

**Is your feature request related to a problem? Please describe.**
When using the "memory" cache and the `@UseCache` decorator, the `refreshThreshold` option does not work because [cache-manager@^4](https://github.com/node-cache-manager/node-cache-manager/blob/4.1.0/lib/stores/memory.js) does not have a `ttl` method. [cache-manager@^5](https://github.com/node-cache-manager/node-cache-manager/blob/master/src/stores/memory.ts#L66) does have the `ttl` method.

**Describe the solution you'd like**
Update the dependencies of `@tsed/platform-cache` to include cache-manager@^5 instead of cache-manager@^4, so that the "memory" cache store works with the `refreshThreshold` option of `@UseCache`.

**Describe alternatives you've considered**
Suggest in the documentation an alternative memory cache package to use that satisfies the above requirements. 

## Acceptance criteria

- [ ]  The "memory" cache store works with the `refreshThreshold` option of `@UseCache`.

Romakita:
- bump cache-manage to v5
- bump @tsed/monorepo-tools to v1.22.6

### DIFF
--- a/packages/platform/platform-cache/src/interfaces/interfaces.ts
+++ b/packages/platform/platform-cache/src/interfaces/interfaces.ts
@@ -1,17 +1,18 @@
 import type {BaseContext} from "@tsed/di";
-import type {Cache, CacheOptions, CachingConfig, Store} from "cache-manager";
+import type {Cache, CachingConfig, Store} from "cache-manager";
+import {FactoryStore} from "cache-manager/dist/caching";
 import type {CacheManager} from "../services/PlatformCache";
 
-export interface PlatformCacheSettings extends CacheOptions, CachingConfig {
+/**
+ * @deprecated Since 2022-11. Use FactoryStore instead.
+ */
+export type PlatformCacheLegacyStoreFactory = {
+  create(...args: any[]): Promise<Store | any> | Store | any;
+};
+
+export type PlatformCacheSettings<Config extends object = any, S extends Store = any> = CachingConfig<Config> & {
   caches?: Cache[];
-  store?:
-    | "memory"
-    | "none"
-    | Store
-    | {
-        create(...args: any[]): Store;
-      }
-    | (() => any);
+  store?: "memory" | Store | PlatformCacheLegacyStoreFactory | FactoryStore<S, Config>;
   max?: number;
   keyResolver?: (args: any[], ctx: BaseContext) => string;
 
@@ -20,7 +21,7 @@ export interface PlatformCacheSettings extends CacheOptions, CachingConfig {
    * otherwise they will be ignored.
    */
   [key: string]: any;
-}
+};
 
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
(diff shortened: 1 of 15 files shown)


packages/platform/platform-cache/src/interfaces/interfaces.ts
```diff
@@ -1,8 +1,8 @@
 import type {BaseContext} from "@tsed/di";
-import type {Cache, CacheOptions, CachingConfig, Store} from "cache-manager";
+import type {Cache, CachingConfig, Store} from "cache-manager";
 import type {CacheManager} from "../services/PlatformCache";
 
-export interface PlatformCacheSettings extends CacheOptions, CachingConfig {
+export type PlatformCacheSettings<Config = any> = CachingConfig<Config> & {
   caches?: Cache[];
   store?:
```
petermenkeepsilon: It looks like cache-manager@^5 removed the both the `'none'` and `{ create(...args: any[]): Store }` types as input to the `caching` method ([source](https://github.com/node-cache-manager/node-cache-manager/blob/master/src/caching.ts#L44-L60)).

You could map the `create` function to a `FactoryStore` easily.
`'none'` used to be its own store type ([source](https://github.com/node-cache-manager/node-cache-manager/blob/4.1.0/lib/stores/none.js)), but has been removed in v5

packages/platform/platform-cache/src/services/PlatformCache.ts
```diff
@@ -66,7 +67,7 @@ export class PlatformCache {
     }
   }
 
-  wrap<T>(key: string, fetch: () => Promise<T>, options?: CachingConfig): Promise<T> {
+  wrap<T>(key: string, fetch: () => Promise<T>, options?: CachingConfig<any>): Promise<T> {
```
petermenkeepsilon: The `CacheManager.wrap` method used below requires a `number` as it's third parameter, not a config

packages/platform/platform-cache/src/services/PlatformCache.ts
```diff
@@ -78,8 +79,9 @@ export class PlatformCache {
     return deserialize(this.cache?.get<T>(key), options);
   }
 
-  async set<T>(key: string, value: any, options?: CachingConfig): Promise<T | undefined> {
-    return this.cache?.set<T>(key, value, options);
+  async set<T>(key: string, value: any, options?: CachingConfig<T>): Promise<T | undefined> {
+    this.cache?.set(key, value, options?.ttl);
```
petermenkeepsilon: `set` can return a promise:
```suggestion
    await this.cache?.set(key, value, options?.ttl);
```

Romakita: true!

Romakita: my bad ^^

packages/platform/platform-cache/src/services/PlatformCache.ts
Romakita: done :)

packages/platform/platform-cache/src/interfaces/interfaces.ts
Romakita: Ok I can do that to be compatible with legacy code

Romakita: Done :)

Issue #2165, opened 2022-11-17
It is recommended to avoid using the default export of socketio io


sh-winter:
path: `tsed/packages/third-parties/socketio/src/services/SocketIOService.ts`

It is recommended to use '* as SocketIO' to replace the default export, because such writing method will throw an error when using 'rollup' packaging.

Poor English, machine translation is used, thanks for understanding.

sh-winter:
![image](https://user-images.githubusercontent.com/48613217/202355504-8f233002-9ee3-4ce1-bf55-280e01f6e9de.png)
`socket.io` does not provide a default export.

Romakita:
Hum… what the report between Ts.ED and rollup? 

Rollup is a front-end bundler and tsed is a backend framework. I hope you don’t use it to build your code for a backend environment!!!
Because I won’t support it!
See you

sh-winter:
The error report is as follows:
![image](https://user-images.githubusercontent.com/48613217/202365372-91e94332-a806-45e7-a452-a5a8ffe645d1.png)

sh-winter:
> Hum… what the report between Ts.ED and rollip?
> 
> Rollup is a front-end bundler and tsed is a backend framework. I hope you don’t use it to build your code for a backend environment!!! Because I won’t support it! See you

Yes, I understand, but I need to build an extremely compact Docker image because the production environment is very sensitive to the image size

sh-winter:
It is only necessary to make such modifications:
![image](https://user-images.githubusercontent.com/48613217/202366557-67405501-b5eb-4e84-b661-efd7d79d21a5.png)

Romakita:
You are free to make a PR. If the test passed, I'll merge it. 

Rollup support default export, and my tsconfig use esInteropModule so it should be correctly supported. I think they many library that use default export, or commonjs exports so asking all projects to change that will be complicated instead of fixing rollup / tsconfig on your side.

See you

sh-winter:
Thank you Bye

Romakita:
@sh-winter  again your are free to make a PR ;)
It’ll take few minutes for you!
See you

sh-winter:
After making changes, I don't know why the test failed, so I used `patch-package` locally to solve this problem.

Romakita:
The ci can confirm if it’s just a timeout or a real problem. Try to push the change on the repo and create a PR. 

see you ;)
Romain

sh-winter:
[...]

PR #2170, opened 2022-11-21
fix(@tsed/socketio): avoid rollup bundling errors (#2165)

sh-winter:
By default, webpack and babel use the cjs export of socket.io, while rollup uses the esm export; Use aggregate import instead of default import to fix bundling errors caused by this difference.

## Information

| Fix                   | No              |

---

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation


Romakita:
All tests are valid. Let’s go to merge ;)

PR #2251, opened 2023-02-07
feat(schema): allow add groupsName to not generate automatically the groups name

### LINKED ISSUE #2249: Specify the OpenAPI schema name when using the @Groups decorator
## Information

When the `@Groups` decorator is used in a controller, the OpenAPI Specification (OAS) will create schemas with names specific to the used groups.
For example, given this controller
```ts
@Controller('/news')
class NewsController {
    async create(
        @BodyParams()
        @Groups('action.create', 'scope.admin')
        @Required()
        news: News,
    ) {
        // ...
    }
}
```

the OAS will contain a schema named `NewsActionCreateScopeAdmin`.

When we change the groups, the schema will be renamed accordingly. Since our clients are generated from the OAS, a change in the groups will create new types for the clients as well.

To prevent this from happening, we would like to be able to specify the name of the schema instead of having it be generated based on the groups.

You've made a suggestion of possible solutions:
> Maybe something like `@Groups(...).GroupsName()`. or alternative `@Groups("Name", ['groups'])` by introducing another signature

Either solution would be fine.

## Acceptance criteria

When it is possible to specify the name of a schema when using the `@Groups` decorator.

Romakita:
Closes: #2249

### DIFF
--- a/docs/docs/model.md
+++ b/docs/docs/model.md
@@ -734,7 +734,7 @@ to use negation by prefixing the group label with `!`.
 ## Groups Name
 
 By default, Groups decorator generate automatically a name for each model impacted by the given groups list.
-If you use a typed client http generator based on Swagger (OAS3) to generate the client code, this behavior can be a constraint for your consumer when change the group list.
+If you use a typed client http generator based on Swagger (OAS3) to generate the client code, this behavior can be a constraint for your consumer when you change the group list.
 
 ```ts
 import {BodyParams, PathParams} from "@tsed/platform-params";
--- a/docs/docs/model.md
+++ b/docs/docs/model.md
@@ -762,7 +762,7 @@ If you change the groups list by this one:
 ```
 
 The new model name will be `UserCreationSummaryExtra`. This change will break the entire consumer code by removing the `UserCreationSummary` type and giving a new `UserCreationSummaryExtra` type.
-In fact, `UserCreationSummaryExtra` and `UserCreationSummaryExtra` are the same model with more fields!
+In fact, `UserCreationSummary` and `UserCreationSummaryExtra` are the same model with more fields!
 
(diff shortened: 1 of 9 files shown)


docs/docs/model.md
```diff
@@ -731,7 +731,65 @@ You can combine different group labels or use a glob pattern to match multiple g
...
+
+In this example, the Groups annotation `@Groups("creation", "summary") user: User` will generate a new model name `UserCreationSummary`.
+If you change the groups list by this one:
+
+```
+ @Groups("creation", "summary", "extra") user: User
+```
+
+The new model name will be `UserCreationSummaryExtra`. This change will break the entire consumer code by removing the `UserCreationSummary` type and giving a new `UserCreationSummaryExtra` type.
+In fact, `UserCreationSummaryExtra` and `UserCreationSummaryExtra` are the same model with more fields!
```
michael-pxr: You probably meant `UserCreationSummary` and `UserCreationSummaryExtra`:
```suggestion
In fact, `UserCreationSummary` and `UserCreationSummaryExtra` are the same model with more fields!
```

docs/docs/model.md
```diff
@@ -731,7 +731,65 @@ You can combine different group labels or use a glob pattern to match multiple g
 to use negation by prefixing the group label with `!`.
 :::
 
-## Groups class definition <Badge text="6.48.0+" />
+## Groups Name
+
+By default, Groups decorator generate automatically a name for each model impacted by the given groups list.
+If you use a typed client http generator based on Swagger (OAS3) to generate the client code, this behavior can be a constraint for your consumer when change the group list.
```
michael-pxr:
```suggestion
If you use a typed client http generator based on Swagger (OAS3) to generate the client code, this behavior can be a constraint for your consumer when you change the group list.
```

Romakita:
@michael-pxr thanks for the review ;)

PR #2255, opened 2023-02-12
2254 discriminator + deserialize does not appropriately create objects when used with arrays

### LINKED ISSUE #2254: Discriminator + Deserialize does not appropriately create objects when used with Arrays.
I'm attempting to use deserialize with Discriminator support.

I noticed an issue, so i'm going to use the example to demonstrate that problem.

The object we create in the example is below:

```ts
const list = {
  data: [
    {
      type: "page_view",
      value: "value",
      url: "https://url"
    },
    {
      type: "action",
      value: "value",
      event: "event"
    },
    {
      type: "click_action",
      value: "value",
      event: "event"
    }
  ]
};
```
Because `data` is an Array, the TS type should be represented as such.

Example below.

```ts
class DataObject {
   data: Array<ObjectType1 | ObjectType2 | ObjectType3>
}
```

If you copy and paste the full example code (below), the issue should become clear.

Specifically, deserialize does NOT make the internal objects appropriately when used with Array.

Example Code
```ts
import { deserialize } from "@tsed/json-mapper";
import {
  DiscriminatorKey,
  DiscriminatorValue,
  OneOf,
  Property,
  Required,
} from "@tsed/schema";

class Event {
  @DiscriminatorKey() // declare this property a discriminator key
  type: string;

  @Property()
  value: string;
}

class SubEvent extends Event {
  @Property()
  meta: string;
}

@DiscriminatorValue("page_view") // or @DiscriminatorValue() value can be inferred by the class name
class PageView extends SubEvent {
  @Required()
  url: string;
}

@DiscriminatorValue("action", "click_action")
class Action extends SubEvent {
  @Required()
  event: string;
}

@DiscriminatorValue()
class CustomAction extends Event {
  @Required()
  event: string;

  @Property()
  meta: string;
}
/**
STOP HERE, COMMENT / UNCOMMENT TO `data` TO SEE THE PROBLEM
*/
class Tracking {
  // If data is NOT an Array, this works.
  @OneOf(PageView, Action, CustomAction)
  data: PageView | Action | CustomAction;

  // If data IS an Array, this fails
  // @OneOf(PageView, Action, CustomAction)
  // data: Array<PageView | Action | CustomAction>;
}

describe("test", () => {
  it("check", () => {
    const list = {
      // Data is an ARRAY, not a single object so the example on the website is indeed incorrect. 
      data: [
        {
          type: "page_view",
          value: "value",
          url: "https://url",
        },
        {
          type: "action",
          value: "value",
          event: "event",
        },
        {
          type: "click_action",
          value: "value",
          event: "event",
        },
      ],
    };

    const result = deserialize(list, {
      type: Tracking,
    });

    expect(result.data[0]).toBeInstanceOf(PageView);
    expect(result.data[1]).toBeInstanceOf(Action);
    expect(result.data[2]).toBeInstanceOf(Action);
  });
});

```

Romakita:
## Information

| fix | No          |

---

Closes: #2254 

## Todos

- [x] Tests
- [x] Coverage

### DIFF
--- a/packages/specs/json-mapper/src/utils/deserialize.ts
+++ b/packages/specs/json-mapper/src/utils/deserialize.ts
@@ -182,8 +182,8 @@ function mapPropStoreOptions(store: JsonPropertyStore, options: JsonDeserializer
     type: store.computedType
   };
 
-  if (store.schema.isDiscriminator) {
-    itemOpts.type = store.schema.discriminator();
+  if (store.itemSchema.isDiscriminator) {
+    itemOpts.type = store.itemSchema.discriminator();
   }
 
   if (store.schema.hasGenerics) {
--- a/packages/specs/json-mapper/test/integration/discriminator.integration.spec.ts
+++ b/packages/specs/json-mapper/test/integration/discriminator.integration.spec.ts
@@ -1,7 +1,17 @@
 import {Controller} from "@tsed/di";
 import {deserialize, serialize} from "@tsed/json-mapper";
 import {BodyParams, PathParams} from "@tsed/platform-params";
-import {DiscriminatorKey, DiscriminatorValue, getSpec, JsonParameterStore, OneOf, Property, Put, Required, Returns} from "@tsed/schema";
+import {
+  CollectionOf,
+  DiscriminatorKey,
+  DiscriminatorValue,
+  JsonParameterStore,
+  OneOf,
+  Property,
+  Put,
+  Required,
+  Returns
+} from "@tsed/schema";
 
 class Event {
   @DiscriminatorKey() // declare this property as discriminator key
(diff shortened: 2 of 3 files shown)


packages/specs/json-mapper/src/utils/deserialize.ts
```diff
@@ -182,8 +182,8 @@ function mapPropStoreOptions(store: JsonPropertyStore, options: JsonDeserializer
     type: store.computedType
   };
 
-  if (store.schema.isDiscriminator) {
-    itemOpts.type = store.schema.discriminator();
+  if (store.itemSchema.isDiscriminator) {
+    itemOpts.type = store.itemSchema.discriminator();
```
endurance, line 186: oh wow, such a tiny change. What does schema / itemschema represent?

packages/specs/json-mapper/test/integration/discriminator.integration.spec.ts
```diff
@@ -44,9 +54,37 @@ class Tracking {
...
 
+class TrackingWithArray {
+  @OneOf(PageView, Action)
+  data: (PageView | Action)[];
+}
+
 describe("Discriminator", () => {
   describe("deserialize()", () => {
     it("should deserialize object according to the discriminator key (model with property discriminator)", () => {
+      const list = {
```
endurance, line 65: Not super important but the variable name isn't correct. should be like... `trackingObj` or something. :D

endurance:
thank you so much!

packages/specs/json-mapper/src/utils/deserialize.ts
Romakita: To represent schema (json schema) for collection and his item I need to have two instance of JsonSchema to store metadata.

if the property is just a simple type ( not a collection), itemSchema === schema.
If it’s collection, schema will represent the collection part (array, map, etc) and itemSchema the schema of the item (class, or primitive class : number, string).

packages/specs/json-mapper/test/integration/discriminator.integration.spec.ts
Romakita: Totally right I’ll change that ;)

PR #2400, opened 2023-08-01
fix(di): preserve original metadata on @Inject decorator

### LINKED ISSUE #2399: @tsed/di overrides native metadata from 'reflect-metadata'
### Description

I have been actively working on integrating the Automock library (https://github.com/automock/automock) with various DI frameworks, including Ts.ED 🎉 . During the exploration of `@tsed/di` metadata mechanism, I came across a potential limitation concerning the `@Inject` decorator.

Presently, when using the `@Inject` decorator, it overrides the native metadata key of the parameter [see here](https://github.com/rbuckton/reflect-metadata/blob/3aeb98af4030be664a66f49bfd164936e0ba1825/README.md#L142). This behavior can have unintended consequences, particularly when integrating with external libraries like Automock, which heavily relies on accurate metadata to automatically mock class dependencies. See: 

https://github.com/tsedio/tsed/blob/83bc31d70aa58c672df3aea05daab38045b3d704/packages/di/src/common/decorators/inject.ts#L47

To address this concern and ensure seamless integration with any library, I propose the addition of an extra metadata key, named `original:paramtypes`, in the `@tsed/di` infrastructure. This new key will store the original parameter types before applying any decorator modifications. By preserving the original metadata, the `@Inject` decorator can continue to function as intended, while also enabling 3rd party libraries to access the unaltered parameter types when needed.

omermorad:
Closes #2399


packages/di/src/common/decorators/inject.ts
```diff
@@ -40,6 +40,8 @@ export function Inject(token?: TokenProvider | (() => TokenProvider), onGet = (b
       case DecoratorTypes.PARAM_CTOR:
         if (token) {
           const paramTypes = Metadata.getParamTypes(target, propertyKey);
+          Metadata.setParamTypes(target, 'design:source:paramtypes', paramTypes);
```
Romakita: This change won't fix your issue because the paramTypes item are modified by reference.

I suggest to change the code by:

```ts
      const store = Store.from(target)
          const paramTypes = [...(store.get("ctor:deps") || Metadata.getParamTypes(target))];
          const type = paramTypes[descriptor as number];

          paramTypes[descriptor as number] = type === Array ? [token] : token;

          store.set("ctor:deps", paramTypes);
```

In the inject.spec.ts replace all:

```
@Inject(InjectorService) readonly injector: Injector
```
by
```
@Inject(InjectorService) readonly injector: any
```
At this step the test will fail.

To fix that, update the InjectorService.ts in mapInvokeOptions:

```ts
else {
      // useClass
      deps = deps ||  provider.store.get('ctor:deps') || Metadata.getParamTypes(provider.useClass);
      construct = (deps: TokenProvider[]) => new provider.useClass(...deps);
    }
```
Re run test. It should works :)

See you

Romakita (changes requested):

Romakita (approved):

PR #2436, opened 2023-09-14
Fixes In decorator parameter usage snippet

lamoglia:
## Information

| Fix | No          |

---

Fixes snippet with an example of usage of In decorator.

According to https://api-docs.tsed.io/api/specs/schema/types/decorators/operations/In.html, its parameter should be a `JsonParameterTypes` or a plain string indicating where the input should be put in (header, body, path, query, cookie...)

### DIFF
--- a/docs/tutorials/snippets/swagger/endpoint-extra-in-params.ts
+++ b/docs/tutorials/snippets/swagger/endpoint-extra-in-params.ts
@@ -6,7 +6,7 @@ import {CalendarModel} from "./models/Calendar";
 @Controller("/calendars")
 export class CalendarCtrl {
   @Post("/")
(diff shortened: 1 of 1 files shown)


lamoglia:
@Romakita , i'm curious about https://github.com/tsedio/api-docs.tsed.io . It seems auto-generated but I couldn't find any indications about which tool generates it. There are a lot of broken links there (at least) in the `Source` indications, eg. https://api-docs.tsed.io/api/specs/schema/types/decorators/operations/In.html

Romakita:
The api doc reference is directly generated from the source code and comments. I developped tool to generate that. But it’s completly broken ^^.

For the in.html it’s directly related to in.ts file :)

see you

Romakita (approved):

PR #2472, opened 2023-10-11
feat(bullmq): add new bullmq integration for

abenerd:
## Information

| Feature | No          |

---

New [BullMQ](https://bullmq.io) integration for Ts.ED to easily create and dispatch jobs. 

## Usage example

```typescript
import {JobController, JobMethods} from "@tsed/bullmq";

@JobController("example")
class ExampleJob implements JobMethods {
  public handle(payload: {msg: string}) {
    console.info("New message incoming", payload.msg);
  }
}
```

```ts
import {Service} from "@tsed/di";
import {JobDispatcher} from "@tsed/bullmq";
import {ExampleJob} from "./jobs/ExampleJob";

@Service()
class MyService {
  constructor(private readonly dispatcher: JobDispatcher) {}

  public async doingSomething() {
    await this.dispatcher.dispatch(ExampleJob, {msg: "this message is part of the payload for the job"});

    console.info("I just dispatched a job!");
  }
}
```

## Todos

- [X] Tests
- [X] Coverage
- [X] Example
- [X] Documentation


docs/tutorials/bullmq.md
```diff
@@ -0,0 +1,148 @@
...
+    content: ts.ed typescript node.js javascript decorators bullmq
+---
+
+# BullMQ
+
+<Banner src="/bullmq.png" href="https://github.com/taskforcesh/bullmq" height="200" />
+
+## Feature
+
+The [BullMQ](https://bullmq.io) Module for Ts.ED allows you to decorate a class using the `@AsJob` decorator and implement the `Job` interface provided by the module.
```
Romakita: Can you rename AsJob by Job and the Job interface by JobMethods. I already had this problems and by convention I add Methods on interface, to keep the decorator name simple ;)

packages/third-parties/bullmq/src/BullMQModule.ts
```diff
@@ -0,0 +1,52 @@
...
+  @Constant("bullmq")
+  private readonly bullmq: BullMQConfig;
+
+  constructor(
+    private readonly injector: InjectorService,
+    private readonly dispatcher: JobDispatcher,
+    private readonly factory: ProcessorFactory
+  ) {}
+
+  $beforeInit() {
```
Romakita, line 17: There is reason to use $beforeInit instead $onInit ?

packages/third-parties/bullmq/src/BullMQModule.ts
```diff
@@ -0,0 +1,52 @@
...
+            defaultJobOptions: this.bullmq.defaultJobOptions
+          }),
+          hooks: {
+            $onDestroy: (queue) => queue.close()
+          }
+        })
+        .invoke<Queue>(`bullmq.queue.${queue}`);
+    });
+
+    this.injector.getMany<Job>("bullmq:cron").map((job) => this.dispatcher.dispatch(getComputedType(job)));
```
Romakita: good job :D. Now you are a @tsed/di expert !

packages/third-parties/bullmq/src/decorators/AsJob.ts
```diff
@@ -0,0 +1,26 @@
...
+    StoreMerge("bullmq", {
+      name,
+      queue,
+      opts
+    }),
+    Injectable({
+      provide: `bullmq.job.${name}`,
+      type: opts.repeat ? "bullmq:cron" : undefined
+    }),
+    Injectable({
```
Romakita: it's probably an error!! I don't understand why you need to declare the injectable provider twice. You'll have two instance in the same container.

Also, you use a class. Give the provide props to Injectable isn't necessary.

Romakita:
```ts
Injectable({
  type: opts.repeat ? "bullmq:cron" : "bullmq:job"
})
```

packages/third-parties/bullmq/src/decorators/AsJob.ts
```diff
@@ -0,0 +1,26 @@
+import {StoreMerge, useDecorators} from "@tsed/core";
+import {Injectable, ProviderType} from "@tsed/di";
+import {JobsOptions} from "bullmq";
+
+export type JobStore = {
```
Romakita: export this type in new interfaces directory to reuse in JobFactory

packages/third-parties/bullmq/src/processors/ProcessorFactory.ts
```diff
@@ -0,0 +1,20 @@
...
+import {Job} from "../contracts/Job";
+
+@Service()
+export default class ProcessorFactory {
+  constructor(private readonly injector: InjectorService) {}
+
+  public build(): Processor {
+    // eslint-disable-next-line require-await
+    return async (job) => {
+      const j = this.injector.get<Job>(`bullmq.job.${job.name}`);
```
Romakita: Here is the right way to retrieve job based on metadata (and to avoid twice injectable provider in container)

```typescript
const job = this.injector
   .getMany<Job>("bullmq:cron")
   .find(instance => 
      Store.from(classOf(instance)).get<JobStore>("bullmq").name === job.name)

if (!job) {
  throw new Error ('Job instance not found')
}



```

Romakita (changes requested):

Romakita:
Hello @abenerd 
Thanks for this great PR. just small changes ;)

Tell me when it's done!
See you
Romain

packages/third-parties/bullmq/src/BullMQModule.ts
abenerd: My thought process was that I wanted this to be ready for when the caller wanted to interact with the queues and workers themselves.

packages/third-parties/bullmq/src/processors/ProcessorFactory.ts
abenerd: I'm not sure how this would be better, as I would potentially be retrieving dozens of services from the container while I only need one and know exactly the name to retrieve it by. I also removed the double injectable so this should not be a problem anymore.

Romakita: I understand, but create multiple instance of the same class in the container isn't the good way. And getMany isn't a method to create injectable provider. is just a method to get all instance of the same type.
Using injector.get is maybe simple but in fact the cost time execution is nearly the same as `getMany` + `filter()`.

But you can add service to resolve that or add a method in your module:

```ts
@Module()
export class BullMQModule implements BeforeInit {
  getJob(name: string) {
     return this.injector
   .getMany<Job>("bullmq:cron")
   .find(instance => 
      Store.from(classOf(instance)).get<JobStore>("bullmq").name === job.name)
  }
}

// alternative

@Module()
export class BullMQModule implements BeforeInit {
   #cronJobs: Map<string, Job>;

   constructor(@Inject("bullmq:cron") cronJobs: Job[]) {
      this.#cronJob = toMap(cronJobs, (instance) => Store.from(classOf(instance)).get<JobStore>("bullmq").name)
   }

  getJob(name: string) {
     return this.#cronJob.get(name)
  }
}

// usage
 const j = this.injector.get<BullMQModule>(BullMQModule).getJob(name);
```

abenerd: Ah yeah I can see that, thank you for the clarification!

abenerd:
BullMQ uses Redis for storage and I'm having problems mocking it in the tests, I don't really want to add it to the CI when I'm the only one using it, would you be able to help me out here @Romakita?

Romakita (approved):

Romakita:
@abenerd I fixed your tests and I remove the unnecessary factory to keep the logic in the module directly. Also I added the runInContext from @tsed/di to allow the context injection in a Service.

On point that maybe need to be improved is the Job decorator. His name is similar to Job from bullmq, and it'll probably cause conflict if the consumer (worker) need to get the job as second parameter of the handler (e.g.: add added it as second handler parameters in the PR).

Maybe renaming the decorator is necessary. Proposals: JobService or JobController ? 
 
 See you
 Romain

Romakita (approved):

abenerd:
Thank you so much for that @Romakita! And yeah I can see how this could be confusing, I changed the decorator name to `JobController` to imply that this is not a service that the user should inject themselves.

Romakita:
Ready ? @abenerd

abenerd:
Yeah, ready to go @Romakita!

PR #2479, opened 2023-10-18
feat(bullmq): improve dispatch job type for better payload type hinting

abenerd:
## Information

| Feature | No          |

---

Using the generic type `T` when typing the job allows for easier autocompletion when specifying the payload for a job without having to explicitly set the generic parameter. 

With a job like

```ts
import {JobController, JobMethods} from '@tsed/bullmq';

@JobController('example')
export default class ExampleJob implements JobMethods {
  handle(payload: { msg: string }) {
    console.info(payload)
  }
}
```

Instead of having to explicitly declare the generic type like this
```ts
import {JobDispatcher} from "@tsed/bullmq";
import ExampleJob from "../../jobs/ExampleJob";

export class ExampleService {

  constructor(
    private readonly dispatcher: JobDispatcher,
  ) {
  }

  async doStuff() {
    return this.dispatcher.dispatch<ExampleJob>(ExampleJob, {msg: 'Hello World'});
  }
}
```

typescript can now infer the type from the first job parameter. 

```ts
import {JobDispatcher} from "@tsed/bullmq";
import ExampleJob from "../../jobs/ExampleJob";

export class ExampleService {

  constructor(
    private readonly dispatcher: JobDispatcher,
  ) {
  }

  async doStuff() {
    return this.dispatcher.dispatch(ExampleJob, {msg: 'Hello World'});
  }
}
```
CleanShot 2023-10-18 at 09 33 45

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/third-parties/bullmq/src/dispatchers/JobDispatcher.ts
+++ b/packages/third-parties/bullmq/src/dispatchers/JobDispatcher.ts
@@ -8,7 +8,7 @@ export class JobDispatcher {
   constructor(private readonly injector: InjectorService) {}
 
   // eslint-disable-next-line require-await
-  public async dispatch<T extends JobMethods>(job: Type<JobMethods>, payload: Parameters<T["handle"]>[0] = {}): Promise<BullMQJob> {
+  public async dispatch<T extends JobMethods>(job: Type<T>, payload: Parameters<T["handle"]>[0] = {}): Promise<BullMQJob> {
     const store = Store.from(job).get<JobStore>("bullmq");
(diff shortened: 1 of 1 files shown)


abenerd:
Test failed setting up node so I think you'd have to retry it manually @Romakita.

Romakita (approved):

abenerd:
I could also change the commit to be a fix so we don't have to release another minor version.

PR #2485, opened 2023-10-25
feat(exceptions): add base exception classes for 3xx 4xx 5xx

ktalebian:
## Information

| -------------    | --------------- |
| Feature | Yes     |

---

Wraps each of the 3xx, 4xx, and 5xx error with their own base class exception. This allows to filter out by 3xx, 4xx, 5xx error exceptions using these base classes

```ts
const error = new NotFound();

if (error instanceof ClientException) {
  // comes out as true
}
```

## Todos

- [x] Tests
- [x] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/specs/exceptions/jest.config.js
+++ b/packages/specs/exceptions/jest.config.js
@@ -5,7 +5,7 @@ module.exports = {
   ...require("@tsed/jest-config"),
   coverageThreshold: {
     global: {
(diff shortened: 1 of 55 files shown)


Romakita (approved):

Romakita:
@ktalebian perfect ;)

Issue #2355, opened 2023-06-07
[FEAT] Non-nullable properties in requestBody can be `null` and are not coerced into the correct type


markusenglund:
## Information

- **Version:** 7.x
- Packages:

It's possible to send in `null` as the value for string, integer and bool properties. Unexpectedly, this will not cause a validation error and the `null` will not be coerced into the correct type.

## Example

```ts
export class Test {
    @Property()
    string?: string;

    @Integer()
    integer?: number;

    @Property(Boolean)
    boolean?: boolean;

    @ArrayOf(String)
    stringArray?: string[];
}
```

Request body: 
```json
{
  "string": null,
  "integer": null,
  "boolean": null,
  "stringArray": [null]
}
```

Inside the controller:

console.log(test) logs:

```js
 {
    string: null,
     integer: null,
     boolean: null,
     stringArray: [ null ]
   }
```

This is definitely not expected behavior. The problem can be fixed by setting `coerceTypes: false` in the AJV options, however that's not what we want to do.
The problem seems to stem from the fact that the TS.ED code uses `coerceTypes: true` but then validates a deepCloned version of the object and returns the original version of the object that doesn't have the corrected types (https://github.com/tsedio/tsed/blob/production/packages/specs/ajv/src/services/AjvService.ts). Is there a reason for that?

## Acceptance criteria

- [ ] `null` should be correctly case into the correct schema type.

Romakita:
Hello @markusenglund 
Can read the doc please. There a `@Nullable` decorator for that.

See you

markusenglund:
I'm well aware of the `@Nullable` decorator. What I'm trying to express is that the current behavior incorrectly accepts and passes on properties that have the value `null` even though the `@Nullable` decorator is _not_ used.

I'm not trying to get the API to accept `null` - I want it to _not_ accept `null` since it violates the schema definition.

Romakita:
Ok understood. The reason is the coerceType is neccessary to accept number a string field and vice versa. I won’t change that because it’s acceptable to use type coercion in this case.

but i don’t want to use the coercion result performed by ajv because it’s the json-mapper to make it that.

your issue is correct but it’s related to ajv. I tried many config with ajv but this is the best config that match the expected behavior actually.

See you

markusenglund:
[...]

PR #2504, opened 2023-11-02
feat(ajv): add returnsCoercedValues option to keep the coerced value by ajv

### LINKED ISSUE #2497: [BUG] Body Validation not working as expected
### Describe the bug

I'm currently migrating a express application to Ts.ED. I used joi for the validation before and currently try to make AJV work, as i like the direct support & swagger integration.

I have a problem making the following work:

```ts
import { Integer, Required } from '@tsed/schema';

export class Test {
  @Integer()
  @Required()
  value: number;
}
```

```ts
@Post('/some-url')
async someFunction(@BodyParams() body: Test) {
  ...
}
```

Also writing `@BodyParams(Test) ...` does not work.

I don't find an exact documentation for this, except from `Any model used on parameter and annotated with one of JsonSchema decorator will be validated with AJV.` [written here](https://tsed.io/docs/validation.html). So i guess my usage is correct and should work?

Also in my `server.ts` i have:

```ts
import '@tsed/ajv';
```

What does work for me is something like:
```ts
@Post('/some-url')
async someFunction(@BodyParams('value') @Integer() @Required value: number) {
  ...
}
```
but i would like to validate the whole body. Is this possible?

### To Reproduce

Should be quite is to set up, see above.

Because this is also a kind of "should this be even possible" - issue, i dind't create an extra reproducable repo. If you need one, let me know.

### Expected behavior

I would expect the full body to be validated correctly when used as described above. The request should ail with the ajv error.

### Code snippets

_No response_

### Repository URL example

_No response_

### OS

macOS

### Node version

18.17.1

### Library version

7.41.2

### Additional context

_No response_

EinfachHans:
## Information

| Fix | No ?          |

---

closes #2497

With this PR the result of a Validation is correctly returned. The resulted data can be different from the input data, when the validation modifies it, which is a common usage.

This is my first PR to this awesome project. Please let me know if i'm missing something 🤔

## Todos

- [x] Tests
- [x] Coverage
- [x] Example
- [x] Documentation


packages/specs/ajv/src/services/AjvService.ts
```diff
@@ -44,6 +44,8 @@ export class AjvService {
           value: localValue
         });
       }
+
```
Romakita, line 50: We can remove the deepClone / localValue , give the value directly to ajv ;)

it’ll improve the performance!

Romakita (changes requested):

Romakita:
@EinfachHans thanks for the PR. Just a small change to optimize the code ;)

see you

EinfachHans:
@Romakita okay somehow an integration test is now failing. Didn't know why it worked locally first...

Anyways, the test that fails is `Scenario8` in **testBodyParams.ts**, which tests a null model input should return the null model back. I'm unsure about this test, can you have a look?

Romakita:
Yes I take a look. This test is here to check if the null value is keep through after validation and deserialization depending on the endpoint configuration.
This one of the side effect of ajv.
I’ll check if it’s pertinent or not.

Romakita:
Ok I already given information on this issue and why isn't fixable easily:
https://github.com/tsedio/tsed/issues/2355

EinfachHans:
Mhh, so should i close this PR?

Romakita (approved):

Romakita:
@EinfachHans I fixed the problem by introducing a new ajv options `returnsCoercedValues`.

Can you review the latest commits please ;)

Romakita:
This PR will solve your issue and another issue related to the coerced value by ajv! Good news :D

EinfachHans:
@Romakita awesome, Thanks! 🚀
Code makes totally sense and documentation is quite clear or me 👍🏼

Romakita:
Ok fixed issue on my local. It should be ok to merge and release after that ;)

PR #2517, opened 2023-11-08
feat(platform-params): make RawPathParams configurable like is it for PathParams (useValidation, useType, etc...)

### LINKED ISSUE #2515: feat: validate raw path params

**First**: I hope i don't annoy you with all of my tickets. If that is the case, let me know^^

I'm currently trying to achieve something like described in [this](https://tsed.io/docs/pipes.html#async-transformation-use-case) of your docs, **but** i would like to validate the path param before the pipe uses it. 

The documentation says:

> In this example, our pipe need to be called with RawPathParams to work properly, because our pipe return an instance of PersonModel. PathParams call automatically the DeserializerPipe and it's not what we want. This is why we using RawPathParams.

I understand this, but for me it doesn't sound like the validation is a problem here, but the mapping is? If that is the case it would be cool to be enable validation for raw params as well 🤔

In my project i created this decorator:

```
return UseParam({
    paramType: ParamTypes.PATH,
    dataPath: '$ctx.request.params',
    expression,
    useValidation: true
  });
```

so i copy & pasted the `RawPathParams` and added the `useValidation: true` which seems to work.

So i'm wondering if this is something which would make sense to add to the library as well, because i would like to use the validation stuff before the pipe so i can ensure the value is correctly.

Atm `RawPathParams` can only by called by the `expression`. It would be good to be able to call it similar to the `PathParams` also with:

```ts
RawPathParams({expression: 'xy', useValidation: true})
```

_No response_

### Additional context

_No response_

### Acceptance criteria

_No response_

Romakita:
Closes: #2515

### DIFF
--- a/packages/platform/platform-params/src/decorators/pathParams.ts
+++ b/packages/platform/platform-params/src/decorators/pathParams.ts
@@ -81,10 +81,17 @@ export function PathParams(...args: any[]): ParameterDecorator {
  * @operation
  * @input
  */
-export function RawPathParams(expression: string) {
+export function RawPathParams(expression: string): ParameterDecorator;
+export function RawPathParams(options: Partial<ParamOptions>): ParameterDecorator;
+export function RawPathParams(): ParameterDecorator;
+export function RawPathParams(...args: any[]): ParameterDecorator {
+  const {expression, useType, useMapper = false, useValidation = false} = mapParamsOptions(args);
   return UseParam({
     paramType: ParamTypes.PATH,
     dataPath: "$ctx.request.params",
-    expression
+    expression,
+    useValidation,
+    useMapper,
+    useType
   });
 }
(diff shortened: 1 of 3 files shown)


packages/platform/platform-params/src/decorators/pathParams.ts
```diff
@@ -81,10 +81,17 @@ export function PathParams(...args: any[]): ParameterDecorator {
...
+export function RawPathParams(): ParameterDecorator;
+export function RawPathParams(...args: any[]): ParameterDecorator {
+  const {expression, useType, useMapper = false, useValidation = false} = mapParamsOptions(args);
   return UseParam({
     paramType: ParamTypes.PATH,
     dataPath: "$ctx.request.params",
-    expression
+    expression,
+    useValidation,
+    useMapper,
```
EinfachHans, line 94: Unsure about `useMapper` and `useType` here, because if a user would use them, then it would make no sense to use `RawPathParams` over `PathParams`, would it? 🤔

EinfachHans (approved):

packages/platform/platform-params/src/decorators/pathParams.ts
Romakita: yes... I realize, is just an equivalent of PathParams with the correct option XD

Romakita: we will say that this will allow the dev to customize the behavior of the decorator starting from the case where everything is false (and vice versa).

PR #2528, opened 2023-11-15
feat(common): add priority option on middleware decorator

### LINKED ISSUE #2516: [BUG] Multer Middleware executed before auth middleware
### Describe the bug

I'm currently in progress of implementing a multer file upload and just noticed that the MulterMiddleware is executed before the auth middleware.

I debugged into here:

https://github.com/tsedio/tsed/blob/63a7911eebc2f48ed56365b57bd5cc61ee5e7413/packages/platform/platform-router/src/domain/PlatformRouters.ts#L80

And you can see the auth middleware is after the multer one:

Bildschirmfoto 2023-11-07 um 16 28 34

### To Reproduce

My controller is simple:

```ts
@Controller('/admin')
// My Custom Auth Decorator, calling UseAuth underneath as described in docs
@AdminAuth({ permissions: [ Read ] })
export class AdminController {
  ...

  @Put()
  @AdminAuth({ permissions: [ Write ] })
  public async add(@MultipartFile('image') image: any) { ... }

  ...  
}
```

### Expected behavior

The Auth Middleware should be executed before the Multer Middleware to not process & upload the uploaded files's if the user hasn't even the permission to do so.

It should be generally one of the first middlewares imo, so before validations and stuff, but i think that is already the case

### Code snippets

_No response_

### Repository URL example

_No response_

### OS

macOS

### Node version

20.9.0

### Library version

7.43.0

### Additional context

_No response_

Romakita:
Closes: #2516


EinfachHans (approved):
Looks understandable for me. A priority to define the middleware order 👍🏼

Question: What happens to the priority when i override a middleware using `OverrideProvider`?

Only thing that i would add, would be some docs? About the `priority` option in general and about the default priorities set by this project (-10 for **PlatformAcceptMimesMiddleware** and 10 for **PlatformMulterMiddleware**).

PR #2536, opened 2023-11-17
feat(bullmq): extend configuration

### LINKED ISSUE #2534: feat(bullmq): improvements

Hey again,

i'm currently looking into the `bullmq` package and think i found some stuff to optimize. I can do this and create a PR, but before i would like to talk about what i would change:

1. Remove `disableWorker` from config - this is unused and not needed because of `workerQueues`
2. Deprecate `defaultJobOptions` and add `defaultQueueOptions` - currently you can only set the `defaultJobOptions` option for a queue, while a queue itself has more options to configure
3. Add queue specific options - In case you want to configure two queues differently, you can not do this atm. I thought to use a type like `queueOptions: Record<string, QueueOptions>`, where the key should be the queue name?
4. Same goes for workers. You can't configure them atm, so i would add a `defaultWorkerOptions` applied to every worker and `workerOptions` (like in 3.) to configure some workers specifically
5. Also i would like to add some listeners to the workers as described in the bullmq docs like:

```ts
worker.on('completed', (job: Job<DataType, ResultType, NameType>, result: ResultType) => {
  // add some logic...
});

worker.on('failed', (job: Job<DataType, ResultType, NameType> | undefined, error: Error) => {
  // add some logic...
});

worker.on('error', (failedReason: Error) => {
  // add some logic...
});
```

users should be able to add these accordingly to their needs, im unsure about what the best way to handle this is. Any recommendations? 🤔

_No response_

_No response_

### Additional context

_No response_

### Acceptance criteria

_No response_

EinfachHans:
## Information

| Feature|No          |

---

closes #2534 

### What changed?

- Added functionality to `disableWorker` (this option was unused, i would not say it is a breaking change 🤔)
- Deprecated `defaultJobOptions`
- Added `defaultJobOptions`, `queueOptions`, `defaultWorkerOptions` & `workerOptions`
- Deep merge the configuration of queues and workers by the connection, default options and specific options
- Improved typing
- Updated docs

<hr>

I don't know how i can test the option merge correctly 🤔 If needed please guide me on how to get the call args there.

### DIFF
--- a/docs/tutorials/bullmq.md
+++ b/docs/tutorials/bullmq.md
@@ -53,7 +53,7 @@ import "@tsed/bullmq"; // import bullmq ts.ed module
     // All other worker configuration will be ignored
     disableWorker: true,
     // Specify for which queues to start a worker for.
-    // Defaultly for every queue added in the `queues` parameter
+    // When undefined falls back to all queues specified.
     workerQueues: ["default"],
     defaultWorkerOptions: {
       // Default worker options which are applied to every worker
--- a/docs/tutorials/bullmq.md
+++ b/docs/tutorials/bullmq.md
@@ -49,6 +49,9 @@ import "@tsed/bullmq"; // import bullmq ts.ed module
         // Specify additional queue options by queue name
       }
     },
+    // Disable the creation of any worker.
+    // All other worker configuration will be ignored
+    disableWorker: true,
     // Specify for which queues to start a worker for.
     // Defaultly for every queue added in the `queues` parameter
     workerQueues: ["default"],
--- a/docs/tutorials/bullmq.md
+++ b/docs/tutorials/bullmq.md
@@ -12,7 +12,7 @@ meta:
 
 ## Feature
 
-The [BullMQ](https://bullmq.io) Module for Ts.ED allows you to decorate a class using the `@Job` decorator and implement the `Job` interface provided by the module.
+The [BullMQ](https://bullmq.io) Module for Ts.ED allows you to decorate a class using the `@JobController` decorator and implement the `JobMethods` interface provided by the module.
 Repeatable Jobs can also be defined using this decorator.
 
 For more information about BullMQ look at the documentation [here](https://docs.bullmq.io/);
@@ -35,16 +35,32 @@ import "@tsed/bullmq"; // import bullmq ts.ed module
 
 @Configuration({
   bullmq: {
-    queues: ["default", "special"];
+    // Specify queue name's to create
+    queues: ["default", "special"],
     connection: {
       // redisio connection options
-    };
-    // optional default job options
-    defaultJobOptions: {};
-    disableWorker: false;
-    // optionally specify for which queues to start a worker for
-    // in case not all queues should be worked on
-    workerQueues: ["default"];
+    },
+    defaultQueueOptions: {
+      // Default queue options which are applied to every queue
+      // Can be extended/overridden by `queueOptions`
+    },
+    queueOptions: {
+      special: {
+        // Specify additional queue options by queue name
+      }
+    },
+    // Specify for which queues to start a worker for.
+    // Defaultly for every queue added in the `queues` parameter
+    workerQueues: ["default"],
+    defaultWorkerOptions: {
+      // Default worker options which are applied to every worker
+      // Can be extended/overridden by `workerOptions`
+    },
+    workerOptions: {
+      special: {
+        // Specify additional worker options by queue name
+      }
+    }
   }
 })
 export class Server {}
@@ -52,7 +68,7 @@ export class Server {}
 
 ## Define a Job
 
-A job is defined as a class decorated with the `@Job` decorator and implementing the `Job` interface of the `@tsed/bullmq` package
+A job is defined as a class decorated with the `@JobController` decorator and implementing the `Job` interface of the `@tsed/bullmq` package
 
 ```ts
 import {JobController, JobMethods} from "@tsed/bullmq";
@@ -82,7 +98,8 @@ class OtherExampleJob implements JobMethods {
 
 ## Defining a repeating job
 
-Jobs that should be run regularly on a schedule can also easily defined using the `@Job` decorator
+Jobs that should be run regularly on a schedule can also easily defined using the `@JobController` decorator.
+Doing so will automatically dispatch it without any data.
 
 ```ts
 import {JobController, JobMethods} from "@tsed/bullmq";
@@ -146,9 +163,13 @@ class MyService {
   constructor(private readonly dispatcher: JobDispatcher) {}
 
   public async doingSomething() {
-    await this.dispatcher.dispatch(ExampleJob, {msg: "this message is part of the payload for the job"}, {
-      delay: 600_000 // 10 minutes in milliseconds
-    });
+    await this.dispatcher.dispatch(
+      ExampleJob,
+      {msg: "this message is part of the payload for the job"},
+      {
+        delay: 600_000 // 10 minutes in milliseconds
+      }
+    );
 
     console.info("I just dispatched a job!");
   }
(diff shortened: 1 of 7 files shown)


EinfachHans:
Looks like i'm not allowed to add reviewers 🤔 So i do it cheeky in this comment:

- @Romakita 
- @abenerd

packages/third-parties/bullmq/src/contracts/JobMethods.ts
```diff
@@ -1,3 +1,5 @@
-export interface JobMethods {
-  handle(payload: unknown): unknown;
+import {Job} from "bullmq";
+
+export interface JobMethods<DataType = any, ReturnType = unknown> {
```
abenerd:
```suggestion
export interface JobMethods<DataType = unknown, ReturnType = unknown> {
```

I think we should use `unknown` here too.

packages/third-parties/bullmq/src/config/config.ts
```diff
@@ -1,13 +1,51 @@
...
+   */
   queues: string[];
+
+  /**
+   * Default connection to use for queue's and worker's
+   */
   connection: ConnectionOptions;
+
+  /**
+   * @deprecated Use defaultQueueOptions instead. Will be removed in the next major release
```
abenerd, line 15: I like the deprecation, that is great!

packages/third-parties/bullmq/src/config/config.ts
```diff
@@ -1,13 +1,51 @@
-import {type ConnectionOptions, type DefaultJobOptions} from "bullmq";
+import {type ConnectionOptions, type DefaultJobOptions, QueueOptions, WorkerOptions} from "bullmq";
```
abenerd:
```suggestion
import {type ConnectionOptions, type DefaultJobOptions, type QueueOptions, type WorkerOptions} from "bullmq";
```

Best make this also type imports I think.

abenerd (changes requested):
For testing the merging you can probably add mock the injector and verify that the config is properly passed in the queue and worker.

EinfachHans:
@abenerd comments addressed, but i'm not clear how to test the merge of the configurations there 🤔

Any idea?

abenerd:
Currently not on my computer anymore, and probably not until Monday so I can't try out stuff myself but maybe try also mocking the `InjectorService` to later verify that it has been called with the correct arguments. We also do something similar in the `BullMQModule.spec.ts`.

EinfachHans:
yeah i tried it, but that seems to not work. I'm unsure, but i think you can't mock the `InjectorService` for `PlatformTest.create` like this. I think the test in `BullMQModule.spec.ts` only works because it tests to not have been called.

Am i right @Romakita ?

Romakita:
No you don't need to mock injectorService to test the merge configuration. just mock `new Queue` using jest.

It's already the case:

```ts
jest.mock("bullmq", () => {
  return {
    Queue: class {
      public name: string;
      public opts: any;

      constructor(name: string, opts: any) {
        this.name = name;
        this.opts = opts;
      }
      close() {}
    },
    Worker: class {
      public name: string;
      public opts: any;

      constructor(name: string, opts: any) {
        this.name = name;
        this.opts = opts;
      }
      close() {}
    }
  };
});
```

Then use PlatformTest.create({bullmq}) to check you different scenario (using different describe).
```ts
import { PlatformTest } from "@tsed/common";
import { instance, mock, when } from "ts-mockito";

describe("module", () => {

  describe("When configuration is given from default server options", () => {
    let dispatcher: JobDispatcher;
    beforeEach(async () => {
      dispatcher = mock(JobDispatcher);
      when(dispatcher.dispatch(CustomCronJob)).thenResolve();

      await PlatformTest.create({
        bullmq: {
          // add default options
        },
        imports: [
          {
            token: JobDispatcher,
            use: instance(dispatcher)
          }
        ]
      });
    });

    afterEach(PlatformTest.reset);

    /// it
  });

  describe("When configuration is given from default server options and decorators", () => {
    let dispatcher: JobDispatcher;
    beforeEach(async () => {
      dispatcher = mock(JobDispatcher);
      when(dispatcher.dispatch(CustomCronJob)).thenResolve();

      await PlatformTest.create({
        bullmq: {
          // add default options
        },
        imports: [
          {
            token: JobDispatcher,
            use: instance(dispatcher)
          }
        ]
      });
    });

    afterEach(PlatformTest.reset);

    /// it
  });
});

```

EinfachHans:
Test added 👍🏼

abenerd:
You know, now that I've slept on in a bit I don't think we should deprecate the `disableWorker` property, it is more explicit and more easily configurable. Configurations are probably read from the environment and if you enable disabling the worker via the `workerQueues` property you have to differentiate between empty values that indicate falling back to using all defined queues and empty values indicating the workers should be disabled. So while it might be a breaking change it would just implement the actual expected behaviour, I think it might be worth it. What do you think @EinfachHans and @Romakita.

EinfachHans:
Mhh i think `disableWorkers` is not very usefull. Why would you have no worker. But even if you want to, you could still do `workerQueues: []`, couldn't you?

abenerd:
We want to decouple the queue from the rest of our api so we can independently scale the two and should there be a catastrophic failure on the worker it would not tear down the api as well. 
That is also the reason you can specify worker queues instead of always processing jobs from all queues so you can work jobs on certain queues on machines with more RAM or CPU for example.

EinfachHans:
Okay and what should `disableWorkers` do? At the end of the day the same as `workerQueues: []` right? Then i would leave it like this and maybe document that somewhere

abenerd:
Sure, but when you read the config from the environment you have to add extra logic to detect whether the `workerQueues` are empty to disable the worker or empty to work on all queues. And having the property `disableWorker` is also a lot more explicit than just an empty array.

EinfachHans:
Feels for me like having two options doing the same thing 🤔 But i can add it back

EinfachHans:
@abenerd done 😊

abenerd:
Thanks!

docs/tutorials/bullmq.md
```diff
@@ -35,24 +35,43 @@ import "@tsed/bullmq"; // import bullmq ts.ed module
...
+    queueOptions: {
+      special: {
+        // Specify additional queue options by queue name
+      }
+    },
+    // Disable the creation of any worker.
+    // All other worker configuration will be ignored
+    disableWorker: true,
+    // Specify for which queues to start a worker for.
+    // Defaultly for every queue added in the `queues` parameter
```
abenerd:
```suggestion
    // When undefined falls back to all queues specified.
```

packages/third-parties/bullmq/README.md
```diff
@@ -53,16 +53,35 @@ import "@tsed/bullmq"; // import bullmq ts.ed module
...
+    queueOptions: {
+      special: {
+        // Specify additional queue options by queue name
+      }
+    },
+    // Disable the creation of any worker.
+    // All other worker configuration will be ignored
+    disableWorker: true,
+    // Specify for which queues to start a worker for.
+    // Defaultly for every queue added in the `queues` parameter
```
abenerd:
```suggestion
    // When undefined falls back to all queues specified.
```

packages/third-parties/bullmq/src/BullMQModule.ts
```diff
@@ -1,7 +1,8 @@
 import {BeforeInit, DIContext, runInContext} from "@tsed/common";
 import {Constant, InjectorService, Module} from "@tsed/di";
+import {deepMerge} from "@tsed/core";
 import {getComputedType} from "@tsed/schema";
-import {Job, Queue, Worker} from "bullmq";
+import {Job, Queue, QueueOptions, Worker, WorkerOptions} from "bullmq";
```
abenerd:
```suggestion
import {Job, Queue, type QueueOptions, Worker, type WorkerOptions} from "bullmq";
```

I think those could be type imports, we're not actually doing anything with them.

abenerd (commented):
Just a few comments on the documentation otherwise this looks good!

EinfachHans:
After this is merged i will add a new pull request for #2537 😊

abenerd (approved):

PR #2538, opened 2023-11-20
feat(bullmq): add `FallbackJobController`

### LINKED ISSUE #2537: feat(bullmq): add default job handler

At the moment a `JobController` needs to be specified with a explicit job name.

In case you want to use a variable job name, you can not use either the job controller handler or the dispatcher.

A solution could be to add a `DefaultJobController` for every job that is not set up explicit and called for every job where no job controller is found.

An alternative which would be really awesome but harder to implement i guess would be to support some kind of regexes, for example to declare a `@JobController` with a name pattern like `event:*`, which would be called for every event like

- event:123
- event:test
- event:xyz

This could be also be a great addition, but might be overhead to implement, maintain and test.

### Additional context

@abenerd what do you think? 😊 

### Acceptance criteria

_No response_

EinfachHans:
## Information

| Feature | No          |

---

closes #2537 

## What changed ?

- Added a `FallbackJobController` Decorator
- Adjust logic to find a matching job handler:
  1. Search via the job name (as before)
  2. Search a fallback job via the queue name
  3. Search a fallback job without queue name
- Adjusted the Dispatcher to be able to dispatch these kind of jobs (jobs without specific class) as well
- Tests added
- Documentation added
- Add myself as a maintainer


EinfachHans:
@abenerd ✌🏼

Romakita:
@EinfachHans can you rebase your branche from production please ;)

EinfachHans:
@Romakita The branch is up to date with production, can you just "squash and merge" it when it is approved?

packages/third-parties/bullmq/src/index.ts
```diff
@@ -7,4 +7,5 @@ export * from "./config/config";
 export * from "./contracts/JobMethods";
 export * from "./contracts/JobStore";
 export * from "./decorators/JobController";
+export * from "./dispatchers/JobDispatcher.options";
```
Romakita: JobDispatcherOptions.ts

Romakita (changes requested):

packages/third-parties/bullmq/src/dispatchers/JobDispatcher.ts
```diff
@@ -1,25 +1,51 @@
...
+      throw new Error(`Queue(${queueName}) not defined`);
     }
 
-    return queue.add(store.name, payload, {
-      ...store.opts,
-      ...options,
+    return queue.add(name, payload, {
+      ...defaultJobOptions,
+      ...options
     });
```
abenerd: Why the change in signature and behavior here? This has nothing to do really with the fallback controller.

Apart from that I also don't really like opening that up to more arbitrary arguments, I think one of the strong suits here is that it is tightly coupled with the declared job controller so that adding invalid jobs or having invalid payloads is harder to achieve. If you want more control over how to dispatch a job you can inject the Queue itself by its name `bullmq.queue.default` / `bullmq.queue.my-awesome-queue`.

packages/third-parties/bullmq/src/BullMQModule.ts
```diff
@@ -74,7 +74,17 @@ export class BullMQModule implements BeforeInit {
...
+    if (job) {
+      return job;
+    }
+
+    const fallbackJobForQueue = this.injector.get<JobMethods>(`bullmq.fallback-job.${queueName}`);
+    if (fallbackJobForQueue) {
+      return fallbackJobForQueue;
+    }
+
+    return this.injector.get<JobMethods>("bullmq.fallback-job");
```
abenerd: Does it maybe make sense to retrieve the fallback handler explicitly so we can change how we add the payload, like make the queue, and job name more accessible.

packages/third-parties/bullmq/src/dispatchers/JobDispatcher.ts
EinfachHans: Mhh i like the fact that users would be to dispatch all kind of jobs via the dispatcher service. I also documented that this way of dispatching only makes sense when a fallback controller is defined.

packages/third-parties/bullmq/src/BullMQModule.ts
EinfachHans: thought about the same thing, but the job itself is already pushed into the handler and within the job you have the job name and the queue name 🤔

packages/third-parties/bullmq/src/dispatchers/JobDispatcher.ts
abenerd: I think the advantage of the dispatcher is that it integrates so well with the already defined jobs. If you want more control you can reach for the queue from the container, which is already possible. This can, and probably should, be documented better that this is possible and how to do it.
Keeping this out of the dispatcher would also keep it cleaner and easier to maintain.

packages/third-parties/bullmq/src/BullMQModule.ts
abenerd: Yeah that is a fair point.

packages/third-parties/bullmq/src/dispatchers/JobDispatcher.ts
EinfachHans: This advantage is not lost, it still works the same as before, it's also extended. What do you say to create a second method for this?

abenerd: I'm still not quite sure about this. I also don't really see the use case here actually, you allow users to dispatch jobs that possibly do not exist for control that they already have.

EinfachHans: In my case i don't have strict names, they can be dynamic for some reason. That was the reason for the fallback controller in general.

Also i would prefer using the dispatcher instead of injecting the queue directly, to have it same across all jobs 🤔

abenerd: Is it not possible to add the dynamic part of the name to the payload? I feel like this is a very special use case that would add a lot of logic to be supported.
Maybe you can write your own dispatcher in your own codebase to dispatch those kinds of jobs, then you would still have a central place where the dispatch logic is handled.

EinfachHans: In my case i use the name because bullmq's `getRepetableJobs` doesn't return the payload of each job (for performance reasons), and i need to find a specific repeatable job, for what i use the name - thats the solution i found online.

I dont think that dynamic names are a special use case 🤔

abenerd: Yeah fair. But let's still move the logic to another method to keep the methods themselves as simple as possible without having to do all the type juggling. Like a `dispatchHeadlessJob` that accepts the name as a string instead of the job type.

Romakita: You can do that to keep the dispatcher implementation simple:

```ts
  public async dispatch(job: Type | JobDispatcherOptions | string, payload: any, opts: JobsOptions = {}): Promise<BullMQJob> {
    const {queueName, name, options} = this.resolveDispatchArgs(job, opts);


    const queue = this.injector.get<Queue>(`bullmq.queue.${queueName}`);

    if (!queue) {
      throw new Error(`Queue(${queueName}) not defined`);
    }

    return queue.add(name, payload, options);
  }

  private resolveDispatchArgs(job: Type | JobDispatcherOptions | string, payload: any, options: JobsOptions = {}) {
    let queueName: string;
    let name: string;
    let defaultJobOptions: JobsOptions | undefined;

    const queue = this.injector.getMany<Queue>("bullmq:queue").find((queue) => queue.name === store.queue);
    if (typeof job === "function") {
      // job is passed as a Type
      const store = Store.from(job).get<JobStore>("bullmq");

      return {
        queueName: store.queue,
        name: store.name,
        options: {
          ...options,
          ...store.options
        }
      }
    }

    if (typeof job === "object") {
      // job is passed as JobDispatcherOptions
      const {queue: queueName, name, ...opts} = job;   // need interface JobDispatcherOptions extends Record<string, any> {}
      /// alternative const {queue: queueName, name, options: opts} = job

      return {
        queueName,
        name,
        options: {
            ...options,
            ...opts
        }
      }
    }

    // job is passed as a string
    return {
      queueName: "default",
      name: job,
      options
    }
  }
```

Romakita:
@EinfachHans For the review and keep commit history clean(in most case), it's preferable to rebase your branch instead of using pull/merge commands. 

Here is the difference:
Pull/merge:
Capture d’écran 2023-11-22 à 08 04 53


Rebase mode: `git rebase origin/production`

Capture d’écran 2023-11-22 à 08 04 32
In this case, you commit is after the latest production tags. It's more clear for me and there is no extra commits :)
and I'm quite reassured to see that, because I know that your dev takes the latest changes into account (even if the pull/merge seems to do it too, that doesn't guarantee me that you haven't made a regression)

I just pushed the branch @EinfachHans (so you need to run `git fetch & git rebase` or reset your local branch to your remote).

See you

EinfachHans:
@Romakita But what about "Squash & Merge" - wouldn't that result in the same?

I normally only use that approach to pull & merge into my feature branches and then squash and merge the branch when merging into the main branch. This produces only one commit and i also can't see any branches in my git visualisation log

Romakita:
@Romakita But what about "Squash & Merge" - wouldn't that result in the same?

It works. Just in your case, the branch started before the latest version and from another branch. I can be confusing ^^. 
I known there is two point of view between rebase / pull. But my experience on different project has shown me that the rebase is more clear to keep history clean and less error.

Also, squashing commit isn't always a good solution, because some intermediate commit could be important to understand some changes (if the change isn't directly related to the feature, but was a prerequisite eg: di update, or core update).

packages/third-parties/bullmq/src/dispatchers/JobDispatcher.ts
```diff
@@ -1,25 +1,61 @@
...
+      queueName = "default";
+      jobName = job;
+    }
+
+    return {
+      queueName,
+      jobName,
+      defaultJobOptions
+    };
+  }
```
abenerd, line 60: I was thinking more of a dedicated method to dispatch a job.

EinfachHans: Yeah, that here is what @Romakita suggested and tbh i'm more a fan of that instead of having a second function. By that the implementation is capsuled and more clear

abenerd: Yeah ok, then let's do it that way.

abenerd (approved):

EinfachHans:
@Romakita Can you review? 😊

Romakita (approved):

PR #2543, opened 2023-12-04
feat(bullmq): allow settings custom job ids from within job class

abenerd:
## Information

| Feature| No          |

---

Make it easier for jobs to declare their own job ids. Giving them access to the payload and injected services. 
This also streamlines the job id creation when the job is dispatched in several places.

The added method is optional and therefore should not be a breaking change. 

```ts
import {JobController, JobMethods} from "@tsed/bullmq";

@JobController("example-with-custom-id")
class ExampleJobWithCustomId implements JobMethods {
  public handle(payload: {num: number}) {
    console.info("look at my awesome number: ", payload.num);
  }

  public jobId(payload: {num: number}): string {
    return `very realistic job id #${payload.num}`;
  }
}
```

## Todos

- [X] Tests
- [X] Coverage
- [X] Example
- [X] Documentation


Romakita (approved):

abenerd:
Why'd you add the `WithJobId` interface again, this is not used anywhere and could lead to confusion with consumers. @Romakita

Romakita:
@abenerd  I haven't added something ^^. I just rebased the branch to fixed the integration test... It haven't solved the problem. also the fail isn't related of your branch. I missed this issue from the last merge (using node v20). I work to fix that ;)

abenerd:
Ah thanks for the clarification, I'll force push again then to remove the interface.

Romakita:
@abenerd  Merged ;)

abenerd:
Thanks!

PR #2551, opened 2023-12-12
feat(di): allow controllers to be scoped to environment

abenerd:
## Information

| Feature | No          |

---

Allow controllers to be scoped to an environment.

```ts
import {Get} from "@tsed/schema";
import {Controller} from "@tsed/di";

@Controller({
  path: "local",
  environments: ["development"]
})
export class LocalCtrl {
  @Get()
  index(): string {
    return [];
  }
}
```

## Todos

- [X] Tests
- [X] Coverage
- [X] Example
- [X] Documentation

### DIFF
--- a/docs/docs/controllers.md
+++ b/docs/docs/controllers.md
@@ -105,6 +105,13 @@ See the following example:
 
 In order to avoid such side-effects, simply move `findAll()` method above `findOne()`.
 
+### Environment scoped controllers
+
+Sometimes you have controllers that you want to only serve locally for development purposes, for this you can specify the environments in which they should be run.
+These environments are then checked against the `NODE_ENV`
+
+<<< @/docs/snippets/controllers/development-only-controller.ts
+
 ## Request
 
 ### Decorators
--- a/docs/docs/snippets/controllers/development-only-controller.ts
+++ b/docs/docs/snippets/controllers/development-only-controller.ts
@@ -1,0 +1,13 @@
+import {Get} from "@tsed/schema";
+import {Controller} from "@tsed/di";
+
+@Controller({
+  path: "local",
+  environments: ["development"]
+})
+export class LocalCtrl {
+  @Get()
+  index(): string {
+    return [];
+  }
+}
--- a/packages/di/src/common/decorators/controller.spec.ts
+++ b/packages/di/src/common/decorators/controller.spec.ts
@@ -8,7 +8,7 @@ class Test {}
 class Dep {}
 
 describe("@Controller", () => {
-  afterAll(() => {
+  afterEach(() => {
     GlobalProviders.delete(Test);
   });
   it("should register a controller with his path and Dependency", () => {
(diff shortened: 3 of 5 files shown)


Romakita:
Hello @abenerd 
it's a really good idea ! But your option is limited to the controller decorator and it would be great to have it for any kind of provider :). I suggest you to perform the filter implementation on DI level directly. 

So to do that:

- Edit the Provider model and add the environments property: https://github.com/tsedio/tsed/blob/production/packages/di/src/common/domain/Provider.ts
- Edit ProviderOpts and add the environments property: https://github.com/tsedio/tsed/blob/production/packages/di/src/common/interfaces/ProviderOpts.ts
- Filter mounted controller: https://github.com/tsedio/tsed/blob/production/packages/di/src/common/utils/resolveControllers.ts#L50

```ts
export function resolveControllers(settings: Partial<TsED.Configuration>): TokenRoute[] {
  const providers = lookupProperties.flatMap((property) => getTokens(settings[property]));
  
   /// get env from settings
  const env = settings.env

  /// filter this list
  return [...resolveRecursively(providers), ...providers].filter((provider) => !!provider.route) as any[];
}
```

- Update createContainer implementation to take a env parameter to filter providers: 
   - https://github.com/tsedio/tsed/blob/79b855d1374c18a86b39653381fe162d7413f3c8/packages/di/src/common/services/InjectorService.ts#L314C1-L314C41
   - https://github.com/tsedio/tsed/blob/79b855d1374c18a86b39653381fe162d7413f3c8/packages/di/src/node/services/DITest.ts#L54

Implementation: https://github.com/tsedio/tsed/blob/production/packages/di/src/common/utils/createContainer.ts#L5

```ts
export function createContainer(rootModule?: Type<any>) {
 // GlobalProviders.entries() => returns Provider
  const container = new Container(GlobalProviders.entries());

  if (rootModule) { // can be removed because there is no usage of rootModule signature in the code base (I check on tsed/cli also
    container.delete(rootModule);
  }

  return container;
}
```

Suggested:

```ts
import {Env} from "@tsed/core"
export function createContainer(env?: Env) { //or settings: DIConfiguration ?
  /// change the code implementation to 
}
```

To cover all this, we can do a test as follows:

```ts
@Controller({path: "/", environments: [Env.DEV]})
class MyController {
}

describe('Test env', () => {
   describe('Env = DEV', ()=> {
     beforeEach(() => PlatformTest.create({env: Env.DEV})
     afterEach(() => Platform.reset())
     
     it('should return the MyController provider', () => {
         expect(PlatformTest.injector.get(MyController)).toBeInstanceof(MyController)
     })
   })

    describe('Env = PROD', ()=> {
       beforeEach(() => PlatformTest.create({env: Env.PROD})
       afterEach(() => Platform.reset())
     
       it('should return the MyController provider', () => {
          expect(PlatformTest.injector.get(MyController)).not.toBeDefined()
      })
    })
})
```

See you :)

abenerd:
You're right that would be even better!
In your snippets you're using the `Env` Enum, I would actually like to avoid that and use a simple string to give people more control over what their environments are called, names like `staging`, and `super-cute-cats` would not be possible with that. @Romakita

Romakita:
Yes you can use a simple string type instead of enum ;)

PR #2570, opened 2024-01-12
feat(bullmq): support bullmq 5

abenerd:
## Information

| Feature|No          |

---

Support bullmq v5, from what I could see there is nothing that requires changing on our part. 

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/third-parties/bullmq/package.json
+++ b/packages/third-parties/bullmq/package.json
@@ -35,11 +35,11 @@
     "@tsed/eslint": "7.44.1",
     "@tsed/schema": "7.55.0",
     "@tsed/typescript": "7.44.1",
-    "bullmq": "^4.12.3",
+    "bullmq": "^4.12.3 || ^5.1.1",
     "eslint": "^8.12.0",
     "ts-mockito": "^2.6.1"
   },
   "peerDependencies": {
-    "bullmq": "^4.12.3"
+    "bullmq": "^4.12.3 || ^5.1.1"
   }
 }
--- a/.npmignore
+++ b/.npmignore
@@ -18,3 +18,4 @@ jest.config.js
 .gitignore
 .idea
 packages
+*.tsbuildinfo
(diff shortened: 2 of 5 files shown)


packages/third-parties/bullmq/package.json
```diff
@@ -35,11 +35,11 @@
...
     "@tsed/schema": "7.55.0",
     "@tsed/typescript": "7.44.1",
-    "bullmq": "^4.12.3",
+    "bullmq": "^4.12.3 || ^5.1.1",
     "eslint": "^8.12.0",
     "ts-mockito": "^2.6.1"
   },
   "peerDependencies": {
-    "bullmq": "^4.12.3"
+    "bullmq": "^4.12.3 || ^5.1.1"
```
abenerd, line 43: I see how we dropped support for the old agenda version in the agenda package and I would actually also love that but thought this might be a breaking change, but if you're fine with that we can also drop v4 support. @Romakita

Romakita: Hello @abenerd 

I haven't dropped v4 support for agenda ;)
https://github.com/tsedio/tsed/blob/0236c4857add562f0544674d9ef21aef425bee72/packages/third-parties/agenda/package.json#L42

I just changed the peer dependencies flag version for agenda :)

See you

Romakita (approved):

EinfachHans:
@abenerd The removal of the connections is not great. Before i was able to apply specific connection attributes different for queues and workers like:

```ts
defaultQueueOptions: {
    connection: {
      ...redisConfig,
      enableOfflineQueue: false
    },
    defaultJobOptions: {
      removeOnComplete: true,
      removeOnFail: true
    }
  },
  defaultWorkerOptions: {
    connection: {
      ...redisConfig,
      maxRetriesPerRequest: null
    }
  }
```

This now doesn't work anymore. It would be better to mark the connection property as optional here instead of omitting it, what do you think?

Romakita:
Ok we need a small fix!

PR #2572, opened 2024-01-15
feat: ignore *.tsbuildinfo in production builds

abenerd:
## Information

| Feature| No          |

---
As see by the the bullmq update, ignore the `*.tsbuildinfo` files reduces the production build size drastically, so I added the rule to the other projects.
CleanShot 2024-01-15 at 17 51 04
CleanShot 2024-01-15 at 17 52 39

CleanShot 2024-01-15 at 17 52 09

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation


abenerd:
I'm not really sure why the failures appear suddenly, they are also present on `production` locally for me so I don't think this relates to the current change, which should not have an impact on this anyway. @Romakita

Romakita:
Yes the problem is related to a change over a core package, but the CI haven't ran the @tsed/formio package. The repo use lerna with --affected to run each tests packages. But sometimes, the tests aren't executed depending on the commit change. 
Because you have changed all packages, you'd just discover an issue on test :).

I'll fix that on your branch :)
see you

Romakita:
Finally found the reason. Node.js 18.19.0 broke the unit test because global cannot be patched to declare atob/btoa functions in jest.

PR #2580, opened 2024-01-23
Feat bullmq options

Romakita:
## Information

Add feature to  @tsed/bullmq packages:
- add `@InjectQueue` and `@InjectWorker` decorators
- auto discovers queues from `@JobController` decorator


Romakita:
Hello @EinfachHans 
Can you review also the PR please ;)

packages/third-parties/bullmq/src/config/config.ts
```diff
@@ -4,7 +4,7 @@ export type BullMQConfig = {
   /**
    * Specify queue name's to create
    */
-  queues: string[];
+  queues?: string[];
```
abenerd, line 7: Maybe we should also deprecate this option then for when we remove it in the next major version seeing as it is no longer needed?

abenerd (commented):
On first glance this looks good to me, will pull later today and try out with our project to see whether I can find something.

EinfachHans (approved):
LGTM

packages/third-parties/bullmq/src/config/config.ts
Romakita: I kept this option because, sometime we need to declare a queue, let the module to create the Queue instance and use it later with a programmatic code (not through a decorator). But maybe I'm wrong, because I use this module since yesterday :p

abenerd: Ah yeah that is a fair point actually, let's keep it like that!

abenerd (approved):

PR #2662, opened 2024-04-09
fix(di): reimplement imports option correctly on DITest.create()

Romakita:
Closes: #2660

## Information

| Fix | No          |

---

## Usage example

```typescript
import {PlatformTest} from "@tsed/common";
import {registerProvider} from "@tsed/common";

registerProvider({
  provide: "TOKEN",
  useFactory() {
    return {token: "token"};
  }
});

describe("OnInit Invoke", () => {
  beforeEach(() =>
    PlatformTest.create({
      imports: [
        {
          token: "TOKEN",
          use: {token: "test"}
        }
      ]
    })
  );
  afterEach(() => PlatformTest.reset()); 
 //....
})

```

## Todos

- [x] Tests

### DIFF
--- a/packages/di/src/common/services/InjectorService.ts
+++ b/packages/di/src/common/services/InjectorService.ts
@@ -293,7 +293,8 @@ export class InjectorService extends Container {
 
     // allow mocking or changing provider instance before loading injector
     this.settings.imports = this.settings.imports
-      ?.map((meta) => {
+      ?.filter((meta) => meta.token !== InjectorService)
+      .map((meta) => {
         if ("token" in meta && "use" in meta) {
           const {token, use} = meta;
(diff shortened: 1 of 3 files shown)


VictoriqueMoe:
why do we need BOTH `registerProvider` and `imports` i assume only `imports` would be needed, as corrently it is obaslete with `registerProvider`

Romakita:
> why do we need BOTH registerProvider and imports i assume only imports would be needed, as corrently it is obaslete with registerProvider

Register provider is normal usage in your code (the real service implementation). Imports is here to override the provider by his mock using PlatformTest.create(). Like is it when you declare an injectable class with `@Injectable`.

Issue #2668, opened 2024-04-23
Override reconnectOnError in tsed ioredis module


ygpark80:
### Is your feature request related to a problem? Please describe.

There is no way to override 'reconnectOnError' since the internal 'reconnectOnError' always overwrites the one in 'redisOptions'.

https://github.com/tsedio/tsed/blob/3269a12a8e77305ce0030ad644cacb84e0da1f3a/packages/orm/ioredis/src/utils/registerConnectionProvider.ts#L67

### Describe the solution you'd like

I've been experiencing issues with ioredis in our AWS Lambda-based ts.ed distribution, specifically receiving 'connection is closed' errors from time to time. Many comments including [this](https://github.com/redis/ioredis/issues/1042) suggest using 'reconnectOnError', but the ts.ed ioredis distribution hard codes this function.

### Describe alternatives you've considered

Perhaps we could change the order to something like the following?

```
          connection = new Redis({
            lazyConnect: true,
            reconnectOnError,
            ...redisOptions
          } as RedisOptions);
```

ygpark80:
Oh, one more thing. I'm not sure if this is going to make sense, but I have copied the registerConnectionProvider to my source base and tested the reconnectOnError, but things didn't change. As I was examining the error, since you are forcing the use of lazyConnect: true and making a connection manually, an error may arise from the connect() function, thus rendering the connection useless. Since our lambda is reusing the instance and thus reusing the broken connection, the error continues to occur when ioredis is used. So, I wrapped the connect() function with a try-catch, since the ioredis internal was using connect().catch(noop) when lazyConnect is not used. Since then, an ETIMEDOUT error has occurred once in a while, but the reconnection logic kicked in, and I haven't seen any 5xx errors since then.

i.e.

```
				try {
					await connection.connect()
					logger.info("Connected to redis database...")
				} catch (error) {
					Sentry.captureException(error)
				}
```

PR #2674, opened 2024-04-24
Add try catch to connect()

ygpark80:
## Information

| Fix | No          |

As I mentioned in [this comment](https://github.com/tsedio/tsed/issues/2668#issuecomment-2072439582), an error can arise during the connect() function and render the component useless. We need to add a try-catch block. In my case, I was getting an ETIMEDOUT error.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling during the connection process in the database management system.

### DIFF
--- a/packages/orm/ioredis/src/utils/registerConnectionProvider.ts
+++ b/packages/orm/ioredis/src/utils/registerConnectionProvider.ts
@@ -69,7 +69,9 @@ export function registerConnectionProvider({provide, name = "default"}: CreateCo
           } as RedisOptions);
         }
 
-        await connection.connect();
+        try {
+          await connection.connect();
+        } catch (e) {}
 
         logger.info("Connected to redis database...");
 


Romakita:
@ygpark80 thanks forvthe PR. Just rework your commit message please (follow the angular commit convention  pls ;))
See you

Romakita:
https://github.com/tsedio/tsed/blob/production/CONTRIBUTING.md

packages/orm/ioredis/src/utils/registerConnectionProvider.ts
```diff
@@ -69,7 +69,9 @@ export function registerConnectionProvider({provide, name = "default"}: CreateCo
           } as RedisOptions);
         }
 
-        await connection.connect();
+        try {
```
Romakita, line 72: But adding this try catch will hide the connection error. I’m not favorable to do add that without loing the error.
Also I prefer to catch the error only if an option is enable. By default, the error must be thrown.

Romakita (changes requested):

packages/orm/ioredis/src/utils/registerConnectionProvider.ts
Romakita: Suggestion:
```ts

        const error = await catchAsyncError(() => connection.connect());

        if (error && !connectionFailSilently) {
          throw error
        }
```

```ts
export interface BaseIORedisConfiguration {
  name?: TokenProvider;
  cache?: boolean;
  connectionFailSilently?: boolean; // add this option
}
```

- add documentation on ioredis.md page to describe this new option (and your usecase can help to understand why this option exists)
- add unit test to cover this new usecase :)

Romakita (changes requested):

packages/orm/ioredis/src/utils/registerConnectionProvider.ts
ygpark80: I'm not a big fan of creating lots of options since ioredis already has tons. lol. If this is used in a serverless environment, not all resources may be available on startup. However, registerConnectionProvider uses lazyConnect: true and tries to connect to Redis during component initialization. But if it fails, the whole server won't start up, which I think is a bigger problem than losing the error.

I think there is another option: do not use lazyConnect: true and leave ioredis as is. ioredis will use the auto-reconnect feature if an incident like mine happens. You don't have to wait until the connection is ready because ioredis already handles (queues) commands internally.

I think Ts.ED should not let people have too many options, as developers should just go ahead and write logic. Adding an option, in my humble opinion, is unnecessary and also confusing.

ygpark80:
> @ygpark80 thanks forvthe PR. Just rework your commit message please (follow the angular commit convention pls ;)) See you

Sorry about that. it has been updated.

packages/orm/ioredis/src/utils/registerConnectionProvider.ts
Romakita: Ok I understand your point, but lazyConnect is here to solve another issue, and it's normal to stop the bootstrap server when the redis connection isn't available, because this problem will cause other issuer later on the runtime.

We can use the lazyConnect option to address your issue and keep the existing behavior as is it.

```ts
        const {name, cache, lazyConnect = true, ...redisOptions} = item;
//...
          connection = new Redis.Cluster(nodes, {
            ...clusterOptions,
            lazyConnect // replace all lazyConnect in the provider
          });
//...
       if (lazyConnect) {
         await connection.connect();
      }
```

I think this is the better option for you and me :D   

See you

ygpark80: Sounds good. I'll do some work sometime this week.

Romakita: let me know when it's ok for you ;)

ygpark80:
Closing — not pursuing this contribution further. Thanks for the review.

PR #2699, opened 2024-05-07
fix(schema): fix custom error message set through .Error() nested decorator

### LINKED ISSUE #2696: [BUG] Custom `.Error` does not work on schemas other than `@Required`
### Describe the bug

Here is a request object:

```ts
class MyRequest {
  @Property()
  @MinLength(10).Error('Title must be at least 10 characters long')
  title: string;
}
```

And the controller

```ts
@Controller("/")
export class HelloWorldController {
  @Inject()
  myService: MyService;

  @Post("/body")
  @Returns(200, String)
  async cache(@BodyParams() body: MyRequest) {
    return 'done';
  }
}
```

I expect to get a custom error message when I make `curl localhost:8083/body -XPOST -H "Content-Type: application/json" -d '{"title":"short"}'` but still get the generic message:

```json
{
  "name": "AJV_VALIDATION_ERROR",
  "message": "Bad request on parameter \"request.body\".\nMyRequest.title must NOT have fewer than 10 characters. Given value: \"foo\"",
  "status": 400,
  "errors": [
    {
      "instancePath": "/title",
      "schemaPath": "#/properties/title/minLength",
      "keyword": "minLength",
      "params": {
        "limit": 10
      },
      "message": "must NOT have fewer than 10 characters",
      "data": "foo",
      "modelName": "MyRequest",
      "dataPath": ".title",
      "requestPath": "body"
    }
  ],
}
```

I tried a few different decorators. It looks like only the `@Required` decorator works.

### To Reproduce

1) Clone https://github.com/ktalebian/tsed-bug
2) `npm i`
3) `npm start`
4) `curl localhost:8083/body -XPOST -H "Content-Type: application/json" -d '{"prop1":"asd"}' | jq .`

### Expected behavior

Expected to see `Title must be at least 10 characters long` message.

### Code snippets

_No response_

### Repository URL example

https://github.com/ktalebian/tsed-bug

### OS

macOS

### Node version

Node 20

### Library version

v7.*

### Additional context

_No response_

Romakita:
Closes: #2696

## Information

| fix | No          |

---

## Todos

- [x] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation


Romakita:
@ktalebian Can review this PR. I need your review over the AJV response error ;)
Thanks a lot!

packages/specs/schema/src/decorators/common/minLength.spec.ts
```diff
@@ -116,5 +116,70 @@ describe("@MinLength", () => {
...
+        instancePath: "/word",
+        keyword: "errorMessage",
+        message: "Title must be at least 10 characters long",
+        params: {
+          errors: [
+            {
+              emUsed: true,
+              instancePath: "/word",
+              keyword: "minLength",
+              message: "must NOT have fewer than 10 characters",
```
ktalebian, line 173: Will this always remain the default error message? Shouldn't this also say `Title must be at least 10 characters long`?

ktalebian: What would happen if you have 

```
class Model {
      @MinLength(10).Error("Title must be at least 10 characters long")
      word: string;

      @MinLength(100).Error("Setence must be at least 100 characters long")
      sentence: string;
    }
```

The response would have a single `message` but the `params.errors` would then have two entries.

I feel the `result.message` should actually be the `@DefaultMessage` decorator that sits on `Model` class, not the individual schemaPath error.

Romakita: Ajv-errors doesn't override the default message. it wrap the original error and add his message.
Look this page, there is all explanation:

https://ajv.js.org/packages/ajv-errors.html

> I feel the result.message should actually be the @DefaultMessage decorator that sits on Model class, not the individual schemaPath error.

This is not the current test @ktalebian ^^. DefaultMsg decorator should works, if there is no other message that match the properties + rules. 

Furthermore, I don't really have control over how the error message came out. Ajv-errors manages everything and I just follow what is indicated in their docs in order to have the expected behavior.

If we really want to have better message management, we would have to write the plugin ourselves to overload the error messages. I clearly don't have the time to do this nor the desire. This module won't be useful to me in a project, because I don't use these error messages in a front-end project.

But if you have time, you can do develop the module. Instead of adding custom keyword (like is do by ajv-errors), the solution could be to override the errors after the validation is done (by matching error.instancePath + errro.keyword).

See you

Romakita: @ktalebian is it ok for you ? (but we haven't a lot of choices XD)

ktalebian: @Romakita sorry, I was on vacation and not really checking my emails. 

Okay that sounds good for now then!

ktalebian (approved):

PR #2711, opened 2024-05-24
feat: add `appendChildrenRoutesFirst` configuration option

### LINKED ISSUE #2706: [BUG] Children Controller Routes should be declared before parent
### Describe the bug

I have the following setup:

```ts
@Controller({
  path: '/parent',
  children: [ChildrenController]
})
export class ParentController {

  @Post(':id')
  public updateUser() { ... }
}
```

```ts
@Controller('/children')
export class ChildrenController {

  @Post()
  public async getAll() { ... }

}
```

when performing a **POST** against `/parent/children`, this is mapped into the post of the `ParentController`, because `children` also matches `:id`.

### To Reproduce

See above, let me know if you need a repo

### Expected behavior

I think children routes should be always declared before the parent routes, does this make sense or brings other problems i don't see atm? 🤔

### Code snippets

_No response_

### Repository URL example

_No response_

### OS

macOS

### Node version

20.12.2

### Library version

7.69.3

### Additional context

_No response_

EinfachHans:
## Information

| Feature |No          |

closes #2706

---

## Description

As described in the original issue the children routes should be appended before the controller routes itself. This PR adds a opt-in solution via a configuration setting. As discussed we should follow the following:

- In next major release (`8.0.0`): Set the default of this value to `true` & deprecate the option
- In the next major version after that (`9.0.0`): Remove the option

## Usage example

```typescript
import {Configuration} from "@tsed/di";

@Configuration({
  appendChildrenRoutesFirst: true
})
export class Server {}
```

## Todos

- [X] Tests
- [X] Coverage
- [X] Example
- [X] Documentation

### DIFF
--- a/packages/di/src/common/interfaces/DIConfigurationOptions.ts
+++ b/packages/di/src/common/interfaces/DIConfigurationOptions.ts
@@ -22,6 +22,12 @@ declare global {
        * Mount controllers
        */
       mount: Record<string, TokenProvider[]>;
+      /**
+       * Append children routes before the controller routes itself
+       *
+       * Will default to true in next major version
+       */
+      appendChildrenRoutesFirst: boolean;
     }
   }
 }
--- a/packages/di/src/common/interfaces/DIConfigurationOptions.ts
+++ b/packages/di/src/common/interfaces/DIConfigurationOptions.ts
@@ -22,12 +22,6 @@ declare global {
        * Mount controllers
        */
       mount: Record<string, TokenProvider[]>;
-      /**
-       * Append children routes before the controller routes itself
-       *
-       * Will default to true in next major version
-       */
-      appendChildrenRoutesFirst: boolean;
     }
   }
 }
--- a/packages/platform/common/src/config/interfaces/PlatformRouterSettings.ts
+++ b/packages/platform/common/src/config/interfaces/PlatformRouterSettings.ts
@@ -1,13 +1,11 @@
-import {JsonMapperGlobalOptions} from "@tsed/json-mapper";
-
-export interface PlatformJsonMapperSettings extends JsonMapperGlobalOptions {}
+export interface PlatformRouterSettings {
+  appendChildrenRoutesFirst?: boolean;
+}
 
 declare global {
   namespace TsED {
     interface Configuration extends Record<string, any> {
-      router?: {
-        appendChildrenRoutesFirst?: boolean;
-      }
+      router?: PlatformRouterSettings;
     }
   }
 }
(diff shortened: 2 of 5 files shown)


packages/di/src/common/interfaces/DIConfigurationOptions.ts
```diff
@@ -22,6 +22,12 @@ declare global {
        * Mount controllers
        */
       mount: Record<string, TokenProvider[]>;
+      /**
```
Romakita: this option should not located here because it's base configuration shared with all tsed package.

You have to add option in a new interface in this directory:

https://github.com/tsedio/tsed/blob/fb19f24be64bebaa715d208fd1b30dba9f3c387a/packages/platform/common/src/config/interfaces

Create a file PlatformRouterSettings with the following content: 

```ts
import {JsonMapperGlobalOptions} from "@tsed/json-mapper";

export interface PlatformJsonMapperSettings extends JsonMapperGlobalOptions {}

declare global {
  namespace TsED {
    interface Configuration extends Record<string, any> {
      router?: {
         appendChildrenRoutesFirst?: boolean;
      }
    }
  }
}
```

Romakita: I don't see where you map the appendChildrenRoutesFirst with the router instance ^^?

Romakita (changes requested):

packages/di/src/common/interfaces/DIConfigurationOptions.ts
EinfachHans: @Romakita added the PlatformRouterSettings file. What do you mean with the second question?

Romakita: @EinfachHans Sorry, I forget that the PlatformRouters is an injectable service ^^. forgot my second answer all is fine.

Romakita (approved):

packages/platform/common/src/config/interfaces/PlatformRouterSettings.ts
```diff
@@ -0,0 +1,13 @@
+import {JsonMapperGlobalOptions} from "@tsed/json-mapper";
+
+export interface PlatformJsonMapperSettings extends JsonMapperGlobalOptions {}
```
Romakita: you can remove this line to fix this error:

```
Error: @tsed/common: src/index.ts(10,1): error TS2308: Module "./config/interfaces/PlatformJsonMapperSettings" has already exported a member named 'PlatformJsonMapperSettings'. Consider explicitly re-exporting to resolve the ambiguity.
```

Romakita (changes requested):

Romakita (approved):

Romakita (approved):

Romakita:
@EinfachHans  Perfect ;)

PR #2725, opened 2024-06-13
Update interfaces.ts

csprance:
Getting the following error when I run build because of this type here. 

I think it might be a bit too broad. Scoped it down to be `Record<string, any>` which seems like a better fit than `object`

```
$ barrelsby --config .barrelsby.json
node_modules/@tsed/platform-cache/lib/types/interfaces/interfaces.d.ts:13:82 - error TS2344: Type 'Config' does not satisfy the constraint 'Record<string, unknown>'.
  Type 'object' is not assignable to type 'Record<string, unknown>'.
    Index signature for type 'string' is missing in type '{}'.

13     store?: "memory" | Store | PlatformCacheLegacyStoreFactory | FactoryStore<S, Config>;
                                                                                    ~~~~~~


Found 1 error in node_modules/@tsed/platform-cache/lib/types/interfaces/interfaces.d.ts:13
```

## Information

| Fix | No          |

---

### DIFF
--- a/packages/platform/platform-cache/src/interfaces/interfaces.ts
+++ b/packages/platform/platform-cache/src/interfaces/interfaces.ts
@@ -10,7 +10,7 @@ export type PlatformCacheLegacyStoreFactory = {
   create(...args: any[]): Promise<Store | any> | Store | any;
 };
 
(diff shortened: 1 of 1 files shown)


Romakita (approved):

csprance:
Sweet! Thanks!

PR #2759, opened 2024-07-22
docs: add example of server-side events in controllers.md

IvaDey:
## Information

| Doc| No          |

---

## Todos

- [x] Documentation

### DIFF
--- a/docs/docs/controllers.md
+++ b/docs/docs/controllers.md
@@ -488,6 +488,47 @@ This example will produce a response with status code 400 and "Not a number" mes
 See our guide on [HttpExceptions to throw customer HttpExceptions](/docs/throw-http-exceptions.md)
 :::
 
+## Server-side events
+
+::: warning
+`compression` middleware should be disabled for correct SSE work
+:::
+
+### Example
+
+```typescript
+import { Controller } from '@tsed/di';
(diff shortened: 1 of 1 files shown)


Romakita:
Hello @IvaDey 

Thanks a lot for this PR :D and I'm sorry for this late reply. I worked on a new feature for Ts.ED based on your PR documentation.

I think we need to provide a more robuste module because the given example is probably not compatible with Koa. So added in SSE example in the doc section doesn't follows the Ts.ED philosophy to provide a cross platform code exemple ^^.
But you couldn't really know that when you were editing the documentation. 

I would to work together on this PR, that add a new `@tsed/sse` module to configure the server automatically and to provide an official Ts.ED API to implement SSE:
https://github.com/tsedio/tsed/pull/2763

This a draft, but I think it will be awesome to have it for all Ts.ED developers. 

While waiting for this PR to move forward, we can just move your code to the tutorial section of the documentation.

We can also indicate that the example is not compatible with Koa and that a solution is being developed to provide official SSE support ;)


See you @IvaDey  and thank again for your involvement!

IvaDey:
Sorry, missed notifications(

But seems this PR is not rellevant more due #2763

Issue #2774, opened 2024-08-01
Option to disable cache control when caching endpoints


pdeveltere:
### Is your feature request related to a problem? Please describe.

Currently when caching an enpoint response with `@UseCache`, the `cache-control` header is set with the ttl of that cache. 

The problem we are facing is that our server side caching is different from the client side caching:
We have an endpoint that returns data based on the accept-language header, which represent the language our app is in. On the server, we cache the response, based on the accept-language header. The problem here is that any http client by default the cache-control header will take precedence over the accept-language , leading to an issue that the response will stay in the previous language even if the apps switched language.

To be able to resolve this issue we ask for the possibility to have more control over this `cache-control` header when configuring the cache 🙏🏽 

### Describe the solution you'd like

A new configuration like this:
```ts
@Configuration({
    cache: {
        disableCacheControl: true;
    } 
})
```

and then change it in the cache interceptor:

```ts
if (!this.disableCacheControl) {
   const calculatedTTL = this.cache.calculateTTL(result, ttl);

   $ctx.response.setHeaders({
     "cache-control": `max-age=${calculatedTTL}`
   });
}
```

PR #2782, opened 2024-08-13
fix(cache): exclude cache-control with new property

pdeveltere:
## Information

| Fix | No          |

## Additional info

Fix for https://github.com/tsedio/tsed/issues/2774

---

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

### DIFF
--- a/packages/platform/platform-cache/src/interceptors/PlatformCacheInterceptor.ts
+++ b/packages/platform/platform-cache/src/interceptors/PlatformCacheInterceptor.ts
@@ -136,11 +136,13 @@ export class PlatformCacheInterceptor implements InterceptorMethods {
 
     const result = await next();
 
-    const calculatedTTL = this.cache.calculateTTL(result, ttl);
+    if (!this.blacklist.includes('cache-control')) {
+      const calculatedTTL = this.cache.calculateTTL(result, ttl);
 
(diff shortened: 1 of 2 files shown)


Romakita:
hello @pdeveltere 
Thanks for the PR. Can you add extra option documentation in cache.md please ?

also to have correct release note you have to write your commit message followings the angular commit convention here:

https://www.conventionalcommits.org/en/v1.0.0-beta.4/

See you :)

packages/platform/platform-cache/src/interceptors/PlatformCacheInterceptor.ts
```diff
@@ -136,11 +139,13 @@ export class PlatformCacheInterceptor implements InterceptorMethods {
 
     const result = await next();
 
-    const calculatedTTL = this.cache.calculateTTL(result, ttl);
+    if (!this.disableCacheControl) {
+      const calculatedTTL = this.cache.calculateTTL(result, ttl);
```
Romakita, line 143: You have an error here. calculatedTTL isn't available after the bloc. You can run the unit test to reproduce issue ;)

Romakita (changes requested):

PR #2806, opened 2024-08-31
feat(di): add AutoInjectable decorator

Romakita:
## Information

| Feature |  No          |

---

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation


docs/docs/providers.md
```diff
@@ -249,7 +249,43 @@ class ProdTimeslotsRepository implements TimeslotsRepository {
 export class Server {}
 ```
 
-## Lazy load provider <Badge text="6.81.0+"/>
+## AutoInjectable <Badge text="7.82.0+" />
+
+AutoInjectable decorator factory that replaces the decorated class' constructor with a parameterless constructor that has dependencies auto-resolved.
```
VictoriqueMoe, line 254: Might be worth mentioning that a AutoInjectable is not managed by the DI context and thus not added to the registry

VictoriqueMoe (approved):

Issue #2965, opened 2025-01-17
[BUG] Several Issues with Jest


ThanatosGit:
### Describe the bug

```
? Choose the target Framework: Express.js
? Choose the architecture for your project: Ts.ED
? Choose the convention file styling: Ts.ED
? Check the features needed for your project Documentation, Testing
? Choose a documentation plugin
? Choose unit framework Jest
? Choose the runtime: Node.js + SWC
? Choose the package manager: PNPM
↓ Write RC files
↓ Write RC files
✔ Initialize package.json
✔ Install plugins
✔ Load plugins
✔ Install plugins dependencies
✔ Generate project files
✔ Install dependencies
✔ Generate barrels files
```
Creates a `package.json` without `jest` itself`.

```json
{
  "name": "jest-test",
  "version": "1.0.0",
  "description": "",
  "scripts": {
    "build": "pnpm run barrels && swc src --out-dir dist -s  --strip-leading-paths",
    "barrels": "barrels",
    "start": "pnpm run barrels && nodemon src/index.ts",
    "start:prod": "cross-env NODE_ENV=production node --import @swc-node/register/esm-register src/index.js"
  },
  "dependencies": {
    "@swc-node/register": "^1.10.9",
    "@swc/cli": "^0.6.0",
    "@swc/core": "^1.10.7",
    "@swc/helpers": "^0.5.15",
    "@tsed/ajv": "^8.4.2",
    "@tsed/barrels": "^6.1.6",
    "@tsed/core": "^8.4.2",
    "@tsed/di": "^8.4.2",
    "@tsed/engines": "^8.4.2",
    "@tsed/exceptions": "^8.4.2",
    "@tsed/json-mapper": "^8.4.2",
    "@tsed/logger": "^7.0.1",
    "@tsed/openspec": "^8.4.2",
    "@tsed/platform-cache": "^8.4.2",
    "@tsed/platform-exceptions": "^8.4.2",
    "@tsed/platform-express": "^8.4.2",
    "@tsed/platform-http": "^8.4.2",
    "@tsed/platform-log-request": "^8.4.2",
    "@tsed/platform-middlewares": "^8.4.2",
    "@tsed/platform-params": "^8.4.2",
    "@tsed/platform-response-filter": "^8.4.2",
    "@tsed/platform-views": "^8.4.2",
    "@tsed/schema": "^8.4.2",
    "ajv": "^8.17.1",
    "body-parser": "^1.20.3",
[...]

PR #2966, opened 2025-01-18
fix(swagger): avoid join error issue by return default string value when import.meta.dirname isn't available

Romakita:
…

fix issue with jest

## Summary by CodeRabbit

- **Chores**
	- Updated directory path resolution mechanism to improve robustness and flexibility when retrieving root directory information.


Romakita:
fix #2965

PR #3074, opened 2025-05-09
fix(platform-http): pass cookie options to clearCookie method

ygpark80:
## Information

| Feature/Fix/Doc/Chore | No          |

---

## Usage example

Usage is the same. It just passes the value of the opts parameter to clearCookie as it is needed to remove cookie properly.

## Todos

- [ ] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cookie removal to ensure cookie options are respected when clearing cookies.
- **New Features**
  - Added support for setting, appending, and clearing cookies with detailed options in response handling.

### DIFF
--- a/packages/platform/platform-http/src/common/services/PlatformResponse.ts
+++ b/packages/platform/platform-http/src/common/services/PlatformResponse.ts
@@ -343,7 +343,7 @@ export class PlatformResponse<Res extends Record<string, any> = any> {
 
   cookie(name: string, value: string | null, opts?: TsED.SetCookieOpts) {
     if (value === null) {
-      this.raw.clearCookie(name);
+      this.raw.clearCookie(name, opts);
       return this;
     }
 
--- a/packages/platform/platform-http/src/common/services/PlatformResponse.spec.ts
+++ b/packages/platform/platform-http/src/common/services/PlatformResponse.spec.ts
@@ -354,7 +354,7 @@ describe("PlatformResponse", () => {
     it("should clear cookie", () => {
       const {res, response} = createResponse();
 
-      response.cookie("token", null, { maxAge: 0, path: "/", httpOnly: true, secure: true, sameSite: "none" });
+      response.cookie("token", null, {maxAge: 0, path: "/", httpOnly: true, secure: true, sameSite: "none"});
       expect(res.headers).toEqual({
         "set-cookie": "token=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; SameSite=None",
         "x-request-id": "id"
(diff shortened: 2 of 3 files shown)


Romakita (approved):

packages/platform/platform-http/src/common/services/PlatformResponse.ts
```diff
@@ -343,7 +343,7 @@ export class PlatformResponse<Res extends Record<string, any> = any> {
 
   cookie(name: string, value: string | null, opts?: TsED.SetCookieOpts) {
     if (value === null) {
-      this.raw.clearCookie(name);
```
Romakita, line 346: Maybe update test can be a good point ;)

Romakita (changes requested):

packages/platform/platform-http/src/common/services/PlatformResponse.ts
ygpark80: Yes, I thought about it but wasn't sure how to test it since I don't know how you guys to tests.
By looking at the API doc however, it was pretty obvious so.. https://expressjs.com/en/api.html#res.clearCookie

If you still want a test case, I'll probably take a look at the current tests.

Romakita: You have a unit test file at the same level ;)

ygpark80: lol, I'll take a look!

ygpark80: i've looked at the test structure and it uses FakeResponse to mock the response but it seems that the actual behavior of the FakeResponse may be different from the real Express (or whatever's underlying http engine) response so not sure if this is what you wanted! I've added three functions from the actual Express Response source code so the expected output should match the actual Express Response behavior.

Romakita: Hi :)

Good job. In my case, I would have simply used a spyOn on `this.raw.clearCookie` to just check the given args to clearCookie method.


```ts
it("should clear cookie", () => {
      const {res, response} = createResponse();
      vi.spyOn(res, 'clearCookie').mockReturnValue();

      response.cookie("token", null, { maxAge: 0, path: "/", httpOnly: true, secure: true, sameSite: "none" });
      expect(res.clearCookie).toHaveBeenCalledWith("token", { maxAge: 0, path: "/", httpOnly: true, secure: true, sameSite: "none" })
    });
```

I didn't think you would reimplement the logic at the FakeResponse level ^^.

Romakita (approved):

PR #3152, opened 2025-09-01
fix(socketio): socket adapter registered twice

AlexDede:
Socket adapter already registered in socket.io

https://github.com/socketio/socket.io/blob/3be6481d9db90a8d6933e2a7dd39fcbb373e30a1/packages/socket.io/lib/index.ts#L325

as it passes through options here:

https://github.com/tsedio/tsed/blob/93dc76e02810d21d4e7230e93f3d75ce66c3b36f/packages/third-parties/socketio/src/services/SocketIOServer.ts#L12

## Information

| Feature/Fix/Doc/Chore | No          |

## Summary by CodeRabbit

- Refactor
  - Removed automatic Socket.IO adapter initialization after server start. Adapters are no longer set from configuration implicitly, which may affect environments relying on this behavior.

- Tests
  - Updated tests to verify that the adapter is not invoked automatically in the HTTP server scenario.

### DIFF
--- a/packages/third-parties/socketio/src/SocketIOModule.ts
+++ b/packages/third-parties/socketio/src/SocketIOModule.ts
@@ -47,10 +47,6 @@ export class SocketIOModule implements AfterListen, OnDestroy {
       this.io.attach(this.httpsServer, {...this.settings});
     }
 
-    if (this.settings.adapter) {
-      this.io.adapter(this.settings.adapter);
-    }
-
     this.getWebsocketServices().forEach((provider) => this.socketIOService.addSocketProvider(provider));
 
     if (!this.disableRoutesSummary) {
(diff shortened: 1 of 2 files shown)


packages/third-parties/socketio/src/SocketIOModule.ts
```diff
@@ -47,10 +47,6 @@ export class SocketIOModule implements AfterListen, OnDestroy {
       this.io.attach(this.httpsServer, {...this.settings});
     }
 
-    if (this.settings.adapter) {
```
Romakita, line 50: But if you remove this line, the adapter won't be registered...

Romakita: ha ok it's registered by the this.io.httpsServer

Romakita (approved):

PR #3456, opened 2026-09-20
fix(testcontainers-mongo): use unique suffixes for test database names

### LINKED ISSUE #3424: [testcontainers-mongo] Random 0–9999 DB name suffix collides across a run, causing E11000 duplicate key failures
https://github.com/tsedio/tsed/blob/db17b891d09b6689975fc8c6d7833af109d928fb/packages/testcontainers/mongo/src/services/ContainerUtils.ts#L67-L69

```markdown
### Information

- **Package:** `@tsed/testcontainers-mongo`
- **Version:** 8.18.0
- **Node:** 20.x
- **MongoDB image:** mongo:7.0.14 (also reproducible on 6.x)

### Description

`getMongoConnectionOptions()` derives the per-test database name from a random integer
in the range **0–9999**:

```js
// packages/orm/testcontainers-mongo/src/services/ContainerUtils.ts
export function getMongoConnectionOptions(id = "", opts = {}) {
    const seed = Math.floor(Math.random() * 10000); // <-- only 10k possible names
    const { dbName = "db-test", ...otherOpts } = opts;
    const url = `${getMongoUrl()}/${dbName}-${seed}`;
    ...
}
```

Meanwhile `startMongoServer()` starts **one** container per process and caches it on
`global[KEY]`, so every `bootstrap()` / `create()` in a run shares that single container:

```js
// ContainerUtils.ts
const container = getEnvironment(KEY) || (await createMongoContainer(image));
```

And `reset()` (the documented teardown) **does not clear data** unless a single collection
name is explicitly passed:

```js
// TestContainersMongo.ts
static async reset(collectionName) {
    if (typeof collectionName === "string") {
        await TestContainersMongo.cleanCollection(collectionName);
    }
    return PlatformTest.reset(); // only disconnects; data is left in place
}
```

**Consequence:** because the name space is only 10,000 and databases are never dropped or
emptied for the run, two `bootstrap()` calls eventually generate the **same** `db-test-<n>`.
A reused database still contains the previous suite's documents, so any test that seeds a
fixed/hard-coded `_id` fails on insert with:

```
MongoServerError: E11000 duplicate key error collection: db-test-<n>.<collection>
index: _id_ dup key: { _id: ObjectId('...') }
```

The probability grows with the number of suites in a run, so this shows up
as **intermittent / flaky** failures — and is far more likely in serial CI runs
(`--no-file-parallelism`, no Vitest retry) than in parallel local runs where a retry happens to
draw a different random name and masks the problem.

### Expected behavior

Each test bootstrap should get an **isolated** database, and/or teardown should reliably clean
the data so a reused name can never leak documents into the next suite.

### Suggested fixes (any one would resolve it)

1. **Guarantee uniqueness instead of low-entropy randomness.** Use a monotonic counter, a
   UUID/crypto suffix, or the worker id (e.g. `process.env.VITEST_POOL_ID` /
   `VITEST_WORKER_ID`, `JEST_WORKER_ID`) so names cannot collide within a run:
   ```js
   const seed = `${process.pid}-${counter++}`; // or randomUUID()
   ```
2. **Make `reset()` clean the whole database by default** (drop or `deleteMany({})` across all
   collections), not only when a single collection name is supplied.
3. At minimum, **document** that `reset()` does not clear data and that DB names are random
   with a tiny keyspace, so consumers know they must clean up themselves.

### Reproduction

With the shared container, force two bootstraps to receive the same suffix (e.g. pin
`Math.random`), seed a document with a fixed `_id` in each, and run them sequentially — the
second insert throws `E11000`. Restoring unique names (or clearing collections on teardown)
makes it pass deterministically.

wrallin:
## Information

| --------------- | ---------------    |
| Fix                      | No                        |

---

Fixes random 0–9999 database name collisions in `@tsed/testcontainers-mongo` that caused `E11000` duplicate key errors. Database names now use `crypto.randomUUID()`.
Fixes #3424

## Todos

- [x] Tests
- [ ] Coverage
- [ ] Example
- [ ] Documentation

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved generated MongoDB database names to use unique identifiers, reducing the risk of name collisions.
  - Preserved configured connection options, including database names, IDs, and direct connection settings.

- **Tests**
  - Added automated coverage for unique database URL generation and connection option handling.
  - Added a test command for the MongoDB package.


wrallin:
pls merge i need that for my project 😭

Romakita (approved):

Issue #3453, opened 2026-09-12
feat(prisma): support the new `prisma-client` generator


Romakita:
## User story

As an `@tsed/prisma` user adopting Prisma's modern `prisma-client` generator,
I want to generate Ts.ED models, repositories, and services from a Prisma Client generated to an explicit directory,
so that I can migrate to recent Prisma versions without retaining `prisma-client-js` as a compatibility dependency.

## Context

Source discussion: https://github.com/orgs/tsedio/discussions/3449

Prisma recommends the `prisma-client` provider and states that `prisma-client-js` will be removed in a future Prisma ORM release. The new provider requires an explicit `output` and generates the client outside `node_modules`.

The workspace already uses Prisma `7.3.0` to develop `@tsed/prisma`, but the integration cannot yet consume this new provider.

## Product decision

Supporting both Prisma client providers is **not** a requirement if the resulting implementation or maintenance burden is significant.

The implementation should first assess the cost of keeping `prisma-client-js` compatibility:

- If it is small and does not introduce fragile conditional behaviour, support both providers during a transition.
- If it is substantial, release a new major version of `@tsed/prisma` that supports **only** `prisma-client`.

In the Prisma-client-only option:

- The migration must be clearly documented, including the required `output` and import-path changes.
- Configuring `prisma-client-js` must fail early with an explicit error explaining that it is unsupported and how to migrate.
- Do **not** silently rewrite or strip `-js` from the configured provider; an automatic transformation could conceal an invalid schema and yield surprising output.

### Current blockers

- `packages/orm/prisma/src/generator.ts` declares `requiresGenerators: ["prisma-client-js"]`, so Prisma still requires the legacy generator.
- `packages/orm/prisma/src/cli/prismaGenerator.ts` exclusively looks up a generator whose provider is `prisma-client-js`; with `provider = "prisma-client"`, that lookup finds nothing and generation fails.
- The client path is derived from `otherGenerators` and the implementation assumes `output` is present. This must be validated against the new provider's required explicit output.
- Generated imports—especially `Prisma.Decimal`—must keep working with a locally generated ESM client and must not silently fall back to `@prisma/client`.

### Areas to validate

- Prisma generator contract: `@prisma/generator-helper`, `@prisma/internals`, `GeneratorOptions`, and DMMF.
- Import resolution for generated models, the `Prisma` namespace, enums, and scalar types.
- ESM output and relative paths, including any required extensions for the selected generation mode.
[...]

PR #3458, opened 2026-09-23
feat(prisma): support the new prisma-client generator

Romakita:
## Summary

Prisma is deprecating `prisma-client-js` in favor of the newer `prisma-client` generator, which requires an explicit `output` outside `node_modules` (see [tsedio/tsed#3453](https://github.com/tsedio/tsed/issues/3453) and source discussion [tsedio/discussions#3449](https://github.com/orgs/tsedio/discussions/3449)).

Cost assessment showed dual support is cheap: no fragile conditional logic is needed except resolving the correct barrel filename per provider (`client.ts` for `prisma-client` vs `index.ts` for `prisma-client-js`), so `@tsed/prisma` now supports both during the transition rather than shipping a breaking major version.

- Accepts `prisma-client` or `prisma-client-js` as the client generator, with an explicit, actionable error when neither is configured or when the configured generator has no `output` (never a silent fallback).
- Resolves the correct barrel file per provider so generated imports (including `Prisma.Decimal`) work with both generators.
- Removed the manifest's `requiresGenerators: ["prisma-client-js"]` hard requirement (Prisma's own AND-only validation can't express "one of"), replaced by our own validation in `onGenerate`.
- Added a `test/prisma-client` fixture; ran a real `prisma generate` against it and type-checked the generated Ts.ED output with `tsc` to confirm imports resolve correctly end-to-end.
- Documented the migration in the package readme and the `prisma.md` tutorial.

## Test plan

- [x] `yarn test` (vitest) — 65/65 passing, including new `prismaGenerator.spec.ts` covering both providers, missing generator, and missing `output`.
- [x] `tsc --build tsconfig.esm.json` — no errors.
- [x] `oxlint` — clean.
- [x] Manual: `prisma generate` against the new `prisma-client` fixture, then `tsc --noEmit` on the generated Ts.ED models — imports resolve, `Prisma.Decimal` resolves correctly.
- [x] Manual: schema with no client generator declared → fails with the explicit, actionable error.
- [x] Regression: existing `prisma-client-js` fixture (`postgres-esm`) regenerates and compiles unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for both Prisma’s newer `prisma-client` generator and the legacy `prisma-client-js` generator.
  * Configuration errors now clearly identify when a supported generator or its required output path is missing.
* **Documentation**
  * Added setup guidance and schema examples for migrating to the newer generator in the Prisma tutorial and package README.


Romakita:
@Hoegsgaard PR ready

